PR #60, opened 2022-11-22
Introduced submodule and hook

daravep:


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "daml"]
+	path = daml
+	url = git@github.com:digital-asset/daml.git
```
soren-da, line 3: for the public read-only repo I suggest we use https here, so we don't require an ssh key for the checkout.

.hooks/post-checkout
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+echo "updating daml submodule"
+git submodule update daml
```
soren-da, line 4: we should also document the initial checkout of the submodule in the readme

soren-da (approved):

.hooks/post-checkout
daravep: should actually work via initDev!

PR #395, opened 2025-10-29
Fix grammar in party_management_service.proto

Budalebah:
## Description

Fixes a grammar error in `party_management_service.proto` documentation comments.

## Changes

Replaces redundant verb usage "the service use will use" with "the service will use" in two locations:
- Line 256: `ExportPartyAcsRequest.wait_for_activation_timeout` field comment
- Line 345: `ClearPartyOnboardingFlagRequest.wait_for_activation_timeout` field comment

## Type of Change

- [x] Bug fix (documentation grammar)
- [ ] New feature
- [ ] Breaking change

## Testing

No code changes, only comment/documentation corrections. No testing required.

PR #397, opened 2025-10-30
Fix grammar in party_management_service.proto

Budalebah:
- Replace 'the service use will use' with 'the service will use' in two locations
- Fixes redundant verb usage in comment documentation


rgugliel-da:
Duplicate of https://github.com/digital-asset/canton/pull/395

PR #438, opened 2026-02-06
feat(external-call): Add proto definitions and data types

trusch:
## Summary

This PR adds the foundational proto definitions and Scala data types for the external call feature. It defines how external call results are represented in transactions and the language AST.

#### `daml_lf2.proto`
```protobuf
enum BuiltinFunction {
  // ... existing values ...
  BUILTIN_EXTERNAL_CALL = 74;
}
```

#### `transaction.proto`
```protobuf
message ExternalCallResult {
  string extension_id = 1;
  string function_id = 2;
  bytes config = 3;
  bytes input = 4;
  bytes output = 5;
}

message Node.Exercise {
  // ... existing fields ...
  repeated ExternalCallResult external_call_results = 18;
}
```

#### `Node.scala`
- Added `externalCallResults: Seq[ExternalCallResult]` to `Exercise` node
- Added `ExternalCallResult` case class

#### `Ast.scala`
- Added `BEExternalCall` to `BuiltinFunction` sealed trait

#### `LanguageFeatureGen.scala`
- Added feature flag for external calls

#### `Transaction.scala`
- Updated transaction traversal to include external call results

#### `SerializationVersion.scala`
- Added support for serializing external call results

#### `Typing.scala`
- Added type checking for `BEExternalCall` builtin

#### `DecodeV2.scala`
- Added decoder for `BUILTIN_EXTERNAL_CALL`

## Testing

- `TransactionSpec.scala` - Transaction with external calls
- `ValueGenerators.scala` - Generators for external call results

### DIFF
--- a/community/daml-lf/archive/src/main/protobuf/com/digitalasset/daml/lf/archive/daml_lf2.proto
+++ b/community/daml-lf/archive/src/main/protobuf/com/digitalasset/daml/lf/archive/daml_lf2.proto
@@ -390,6 +390,8 @@ enum BuiltinFunction {
   TYPE_REP_TYCON_NAME = 3011; // *Available in versions >= 2.dev*
 
   //SECP256K1_VALIDATE_KEY = 4001; // REMAINS HERE TEMPORARELY TO PARSE CHECKED IN DARS
+
+  EXTERNAL_CALL = 5001; // *Available in versions >= 2.dev*
 }
 
 // Builtin literals
--- a/community/daml-lf/transaction/src/main/protobuf/com/digitalasset/daml/lf/transaction.proto
+++ b/community/daml-lf/transaction/src/main/protobuf/com/digitalasset/daml/lf/transaction.proto
@@ -74,6 +74,8 @@ message Node {
     repeated string observers = 8;
     // No listed authorizers indicates default authorizers (signatories + actors)
     repeated string authorizers = 1001; // *since version dev*
+    // Results of external calls made during this exercise
+    repeated ExternalCallResult external_call_results = 1002; // *since version dev*
   }
 
   message Rollback {
@@ -115,4 +117,17 @@ message FatContractInstance {
 message KeyWithMaintainers {
   bytes key = 1;
   repeated string maintainers = 2; // the maintainers induced by the key
+}
+
+// Result of an external call made during transaction execution.
+// External calls are deterministic HTTP calls to extension services.
+// Results are stored in the transaction and replayed during validation.
+// *since dev*
+message ExternalCallResult {
+  string extension_id = 1;    // Extension service identifier (e.g., "oracle-price-feed")
+  string function_id = 2;     // Function identifier within the extension
+  string config_hash = 3;     // Hash of the extension configuration for version validation
+  string input_hex = 4;       // Hex-encoded input data sent to the extension
+  string output_hex = 5;      // Hex-encoded output data returned by the extension
+  int32 call_index = 6;       // Index of this call within the exercise node (for ordering)
 }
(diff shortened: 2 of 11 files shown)


community/daml-lf/transaction/src/main/protobuf/com/digitalasset/daml/lf/transaction.proto
```diff
@@ -116,3 +118,16 @@ message KeyWithMaintainers {
...
+// Result of an external call made during transaction execution.
+// External calls are deterministic HTTP calls to extension services.
+// Results are stored in the transaction and replayed during validation.
+// *since dev*
+message ExternalCallResult {
+  string extension_id = 1;    // Extension service identifier (e.g., "oracle-price-feed")
+  string function_id = 2;     // Function identifier within the extension
+  string config_hash = 3;     // Hash of the extension configuration for version validation
+  string input_hex = 4;       // Hex-encoded input data sent to the extension
+  string output_hex = 5;      // Hex-encoded output data returned by the extension
```
paulbrauner-da, line 131: config_hash, input_hex, and output_hex should have type bytes, not string. The daml language doesn't support strings, that's why we use an hex encoding in LF. (We may introduce proper support for bytes in LF one day.) But in the transaction, it is a waste of space to represent them as strings.

community/daml-lf/transaction/src/main/protobuf/com/digitalasset/daml/lf/transaction.proto
```diff
@@ -116,3 +118,16 @@ message KeyWithMaintainers {
...
+  int32 call_index = 6;       // Index of this call within the exercise node (for ordering)
```
paulbrauner-da, line 132: This not needed because repeated is ordered.

community/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Node.scala
```diff
@@ -10,6 +10,31 @@ import com.digitalasset.daml.lf.transaction.BackwardsCompatibilityImplicits._
...
+  * @param outputHex Output data (hex-encoded)
+  * @param callIndex Index of this call within the exercise (for ordering)
+  */
+final case class ExternalCallResult(
+    extensionId: String,
+    functionId: String,
+    configHash: String,
+    inputHex: String,
+    outputHex: String,
+    callIndex: Int,
```
paulbrauner-da, line 31: Same comment as in the protobuf definition: configHash, inputHex and outputHex should be `daml.lf.data.Bytes`. callIndex is not necessary.

community/daml-lf/archive/src/main/protobuf/com/digitalasset/daml/lf/archive/daml_lf2.proto
```diff
@@ -390,6 +390,8 @@ enum BuiltinFunction {
   TYPE_REP_TYCON_NAME = 3011; // *Available in versions >= 2.dev*
 
   //SECP256K1_VALIDATE_KEY = 4001; // REMAINS HERE TEMPORARELY TO PARSE CHECKED IN DARS
+
+  EXTERNAL_CALL = 5001; // *Available in versions >= 2.dev*
```
paulbrauner-da, line 394: LF (the language that the daml compiler produces and that the engine consumes) is separate from the language of the transactions produced by the engine, even if they tend to refer to them as "LF transactions". For that reason, we tend to introduce LF changes and transaction changes in separate PRs. I suggest you split this PR into a transaction one and an LF one.

When you modify LF you also need to:
  - Modify LF syntax test parser in `com.digitalasset.daml.lf.testing.parser` and its spec.
  - Add a trivial test for typing in `TypingSpec` (which uses the parser above)
  - Add it to `Builtin_2.dev_.lf`

paulbrauner-da (changes requested):
Overall looks good but please address comments.

PR #486, opened 2026-04-16
Parallelize per-member ECIES encryption and commitment computation (#494)

elliottdehn:
Closes #494

## Summary

Parallelize per-member ECIES encryption, per-participant commitment hashing, per-signature verification, and RunningCommitments updates.

## Impact

**Encrypt-side latency (submitter):**
| Scenario | Before | After | Speedup |
|---|---|---|---|
| 2-party contract (4 ECIES) | 1.6ms | 0.8ms | 2x |
| CC transfer, 20 SVs (44 ECIES) | 16ms | ~3.4ms | 4.7x |

**Total CPU unchanged** — same work, distributed across cores. Benchmarked: 4.7x-8.6x on the ECIES hot path (200 iterations, JDK 21, 14 cores).

**Commitment computation**: `computeCommitmentsPerParticipant` parallelized via ForkJoinPool. Runs every reconciliation interval (~5s). With 50+ counter-participants, sequential → parallel.

**RunningCommitments.update**: groups contract activations/deactivations by stakeholder set, processes independent groups in parallel. Each group has its own LtHash16 accumulator.

## Relation to other PRs

- **Independent of #487** (decrypt-side optimization). Both can land separately.
- **Superseded by #489** for the encrypt path — if ECDH is precomputed in the background, there's nothing left to parallelize on the critical path. However, #486 still helps when the precompute cache is cold (first transaction with new recipients).
- **Superseded by #490** for warm channels. Same reasoning.
- **The commitment and RunningCommitments changes are independent** of all other PRs and always provide value.

## Test plan

- [ ] Existing `EncryptedViewMessageFactory` tests pass
- [ ] `SimplestPingIntegrationTest` passes
- [ ] Microbenchmark: 4.7x speedup for 10 recipients, 8.6x for 20

<details>
<summary>Microbenchmark results</summary>

| Recipients | Sequential (p50) | Parallel (p50) | Speedup |
|---|---|---|---|
| 2 | 0.9ms | 0.5ms | **1.9x** |
| 5 | 2.0ms | 0.5ms | **4.2x** |
| 10 | 4.0ms | 0.9ms | **4.7x** |
| 20 | 8.0ms | 0.9ms | **8.6x** |

</details>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #487, opened 2026-04-16
Bridge session key encryption cache to decryption cache (#495)

### LINKED ISSUE #495: Session key encryption cache not bridged to decryption cache
When a submitter encrypts session key randomness for recipients, the plaintext randomness is not stored in the decryption cache. When the same participant later confirms its own transaction, it performs a redundant ECIES decrypt on data it just encrypted. With 20 SVs this wastes ~8ms per self-confirmed transaction.

PR: #487

elliottdehn:
Closes #495

## Summary

When a submitter encrypts session key randomness for N recipients, also store `(encryptedRandomness → plainRandomness)` in the decryption cache. When the same participant later confirms its own transaction, the ECIES decrypt is a cache hit — no asymmetric crypto needed.

## Impact

**Decrypt-side savings (confirming participant, self-confirmed transactions only):**
|---|---|---|---|
| 2-party self-confirm | 2 ECIES decrypts (0.8ms) | 0 | 0.8ms |
| CC transfer, 20 SVs, self-confirm | 20 ECIES decrypts (~8ms) | 0 | **~8ms** |

**Only applies to self-confirmed transactions** — when the submitting participant is also a confirmer (which is the common case: the submitter is always an informee of its own transaction).

Does NOT help other confirmers (SVs confirming someone else's transaction). Their decrypt cost is unchanged.

**7 lines of code.** Zero risk — just populating a cache that already exists.

## Relation to other PRs

- **Stacks with #489 and #490**: those optimize the **encrypt** side (submitter). This optimizes the **decrypt** side (confirmer). Independent code paths, both provide value.
- **Independent of #486 and #488**: those parallelize/batch the encrypt side. This is decrypt-only.
- **This is the single highest-impact-per-line-of-code change** in the series.

## Test plan

- [ ] Existing session key cache tests pass
- [ ] Session key decryption cache hit rate increases for self-confirmed transactions

### DIFF
--- a/community/base/src/main/scala/com/digitalasset/canton/store/SessionKeyStore.scala
+++ b/community/base/src/main/scala/com/digitalasset/canton/store/SessionKeyStore.scala
@@ -62,8 +62,19 @@ sealed trait ConfirmationRequestSessionKeyStore {
 
   private[canton] def saveSessionKeysInfo(
       toSave: Map[RecipientGroup, SessionKeyInfo]
-  ): Unit =
+  ): Unit = {
     sessionKeysCacheRecipients.putAll(toSave)
+    // Pre-populate the decryption cache so that when this participant acts as a
+    // confirmer for its own submission, the ECIES decrypt is a cache hit.
+    // Without this, the submitter encrypts session key randomness for each recipient
+    // and then the confirmer path decrypts the same bytes — a redundant ECIES operation.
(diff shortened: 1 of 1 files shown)


soren-da:
The decryption cache pre-population makes sense, I'll merge it.

soren-da (approved):

PR #488, opened 2026-04-16
Batch ECIES encryption with shared ephemeral keypair (#496)

### LINKED ISSUE #496: ECIES generates redundant ephemeral keypairs for multi-recipient encryption
When encrypting session keys for N recipients, each call to encryptWith generates an independent ephemeral EC keypair. EC keygen is 16.5% of per-recipient ECIES cost. Sharing a single ephemeral keypair via IESEngine saves ~13% for 20 recipients.

PR: #488

elliottdehn:
Closes #496

## Summary

When encrypting session key randomness for multiple recipients, generate ONE ephemeral EC keypair and reuse it for all ECDH key agreements (via BouncyCastle's `IESEngine`). Saves the EC key generation cost (~0.077ms, 16.5% of ECIES) for N-1 recipients.

## Impact

**Encrypt-side savings (submitter):**
|---|---|---|---|
| 2-party contract | 1.6ms | 1.5ms | 0.1ms (6%) |
| CC transfer, 20 SVs | 16ms | ~14ms | **~2ms (13%)** |
| CC transfer, 40 SVs | 32ms | ~29ms | ~3ms (9%) |

**Modest but free** — zero protocol changes, backward compatible output format.

ECDH key agreement is 77% of per-recipient ECIES cost and cannot be shared (different recipient keys). EC keygen is 16.5% and IS shared by this PR. The savings are real but bounded by keygen being the minority cost.

## Relation to other PRs

- **Superseded by #489** (background precompute): if the entire ECDH is precomputed off the critical path, the keygen savings from batching are already captured. #488 only matters when the precompute cache is cold.
- **Superseded by #490** (channel keys): channel keys eliminate both keygen AND ECDH for warm channels.
- **Independent of #487** (decrypt-side): different code paths.
- **Can land independently** as a low-risk incremental improvement. Provides value as a fallback when #489/#490 caches are cold (first transaction with new recipient groups).

## Test plan

- [ ] Existing ECIES encrypt/decrypt round-trip tests pass
- [ ] Batch-encrypted ciphertexts decrypt correctly with existing Cipher decrypt path

### DIFF
--- a/community/base/src/main/scala/com/digitalasset/canton/crypto/Encryption.scala
+++ b/community/base/src/main/scala/com/digitalasset/canton/crypto/Encryption.scala
@@ -67,6 +67,19 @@ trait EncryptionOps {
       encryptionAlgorithmSpec: EncryptionAlgorithmSpec = encryptionAlgorithmSpecs.default,
   ): Either[EncryptionError, AsymmetricEncrypted[M]]
 
+  /** Encrypts the same message for multiple recipients, sharing a single ephemeral keypair.
+    * This amortizes the EC key generation cost (O(1) keygen instead of O(N)) while performing
+    * O(N) ECDH key agreements. Safe under the multi-recipient DHIES security model.
+    *
+    * Falls back to individual encryptWith calls if the algorithm doesn't support batching.
+    */
+  def batchEncryptWith[M <: HasToByteString](
+      message: M,
+      publicKeys: Seq[EncryptionPublicKey],
+      encryptionAlgorithmSpec: EncryptionAlgorithmSpec = encryptionAlgorithmSpecs.default,
+  ): Either[EncryptionError, Seq[AsymmetricEncrypted[M]]] =
+    publicKeys.traverse(k => encryptWith(message, k, encryptionAlgorithmSpec))
+
   /** Deterministically encrypts the given bytes using the given public key. This is unsafe for
     * general use, and it's only used to encrypt the decryption key of each view.
     */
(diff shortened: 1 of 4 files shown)


elliottdehn:

elliottdehn:
## Benchmark Results

Measured the actual per-step ECIES cost breakdown and the batch savings:

### ECIES Step Breakdown (per recipient, secp256r1)

| Step | Cost | % of total |
|---|---|---|
| ECDH key agreement | 0.361ms | **77.2%** |
| EC keygen (ephemeral) | 0.077ms | 16.5% |
| AES-GCM encrypt | 0.024ms | 5.2% |
| KDF (SHA-256) | 0.005ms | 1.2% |
| **Total** | **0.468ms** | |

### Batch vs Individual Encryption

| Recipients | Individual (p50) | Batch (p50) | Saved | % saved |
|---|---|---|---|---|
| 2 | 0.9ms | 0.8ms | 0.1ms | 11% |
| 5 | 2.0ms | 1.8ms | 0.2ms | 12% |
| 10 | 4.0ms | 3.5ms | 0.6ms | 14% |
| 20 | 8.0ms | 6.9ms | 1.0ms | 13% |
| 40 | 15.9ms | 13.6ms | 2.3ms | 14% |

Correctness verified: all recipients decrypt successfully with the shared ephemeral key.

### Analysis

The keygen savings (~0.077ms per skipped keygen) are real but modest — ECDH key agreement dominates at 77% of per-recipient cost. For a 2-view CC transfer with 20 SVs, this saves ~2ms total. Not transformative by itself, but it's free performance with no protocol changes.

The fundamental O(SV) cost is the ECDH key agreement (0.36ms × N_SV), which cannot be reduced without protocol-level changes (group encryption, threshold encryption, or removing DSO as informee per CIP-0104).

PR #489, opened 2026-04-16
Background ECDH precomputation (#497)

### LINKED ISSUE #497: ECDH key agreement computed on transaction critical path despite being precomputable
ECDH key agreement (77% of per-recipient ECIES cost, 0.36ms each) is computed synchronously during transaction submission. Recipient public keys are known from topology and don't change between transactions. Precomputing ECDH in a background thread moves ~15ms off the critical path for 20-SV CC transfers.

PR: #489

elliottdehn:
Closes #497

## Summary

Precompute ECDH shared secrets with known participants in a background thread. When a transaction needs to encrypt views, the expensive ECDH key agreement (~0.36ms per recipient, 77% of ECIES cost) is already done — only KDF + AES remains on the critical path (~0.03ms).

## Impact

**Encrypt-side latency (submitter, warm cache):**
|---|---|---|---|
| 2-party contract | 1.6ms | ~0.12ms | 1.5ms (93%) |
| CC transfer, 20 SVs | 16ms | **~1.2ms** | **14.8ms (93%)** |

**Total CPU unchanged** — the ECDH work still happens, just in a background thread between transactions instead of on the critical path. Cache warmup takes ~50ms (one background refill cycle).

Each precomputed entry is consumed exactly once (forward secrecy) and automatically refilled in the background.

## Relation to other PRs

- **Stacks with #487** (decrypt-side cache bridge). #489 optimizes encrypt, #487 optimizes decrypt. Both are independent code paths. Combined: **~23ms saved per self-confirmed CC transfer with 20 SVs** (~15ms encrypt + ~8ms decrypt).
- **Supersedes #486 and #488 for the encrypt path**: if ECDH is precomputed, there's nothing left to parallelize (#486) or batch (#488) on the critical path. However, #486's commitment/RunningCommitments parallelism is independent and still valuable. And #486/#488 still help when the precompute cache is cold.
- **Superseded by #490** (channel keys) for warm channels: channel keys achieve ~0.005ms per recipient (HKDF only) vs ~0.03ms (KDF + AES) from precompute. But #489 has **zero protocol changes** while #490 requires a new protobuf enum and wire format. #489 is the safer, faster-to-ship option.
- **No protocol changes. No wire format changes. Pure implementation optimization.**

## Test plan

- [ ] Existing ECIES encrypt/decrypt tests pass (output format unchanged)
- [ ] Precompute cache correctly refills after consumption
- [ ] Forward secrecy: each precomputed entry used exactly once
- [ ] Fallback to full ECIES when cache is cold

### DIFF
--- a/community/base/src/main/scala/com/digitalasset/canton/crypto/SyncCryptoApiParticipantProvider.scala
+++ b/community/base/src/main/scala/com/digitalasset/canton/crypto/SyncCryptoApiParticipantProvider.scala
@@ -599,7 +599,7 @@ class SynchronizerSnapshotSyncCryptoApi(
     MemberType,
     AsymmetricEncrypted[M],
   ]] = {
-    def encryptFor(keys: Map[Member, EncryptionPublicKey])(
+    def encryptForMember(keys: Map[Member, EncryptionPublicKey])(
         member: MemberType
     ): Either[(MemberType, SyncCryptoError), (MemberType, AsymmetricEncrypted[M])] = keys
       .get(member)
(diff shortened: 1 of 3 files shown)


elliottdehn:

PR #490, opened 2026-04-16
Per-participant ECDH channel keys (#498)

### LINKED ISSUE #498: Fresh ECDH per transaction for known participant pairs
Every transaction performs fresh ECDH key agreement per recipient (~0.36ms each), even for participant pairs that have communicated before. Caching the shared secret per pair and deriving per-transaction keys via HKDF reduces the cost to ~0.005ms (72x). Requires new EncryptionAlgorithmSpec for the channel-key ciphertext format.

PR: #490

elliottdehn:
Closes #498

## Summary

New `EncryptionAlgorithmSpec.EcdhChannelAes128Gcm`: caches per-participant-pair ECDH shared secrets and derives per-transaction AES-128-GCM keys via HKDF. After the first ECDH with a recipient (~0.36ms), subsequent encryptions cost ~0.005ms (72x faster).

Full implementation: protobuf enum, encrypt path, decrypt path, channel key store with epoch-based rotation.

## Impact

**Encrypt-side latency (submitter, warm channels):**
| Scenario | Before | After (warm) | Saved |
|---|---|---|---|
| 2-party contract | 1.6ms | ~0.04ms | 1.56ms (97%) |
| CC transfer, 20 SVs | 16ms | **~0.4ms** | **15.6ms (97%)** |
| First contact (cold) | same | same | 0 (full ECDH on first use) |

**This is the highest-impact encrypt-side optimization** in the series, but requires a protocol change (new protobuf enum value).

## Relation to other PRs

- **Stacks with #487** (decrypt-side cache bridge). #490 optimizes encrypt, #487 optimizes decrypt. Independent code paths. Combined for self-confirmed CC transfer: **~24ms saved** (~16ms encrypt + ~8ms decrypt).
- **Supersedes #486, #488, #489 for the encrypt path** when channels are warm. Channel keys eliminate both keygen AND ECDH, making parallelism (#486), batch keygen (#488), and precompute (#489) unnecessary for known recipients.
- **#489 is the code-only alternative** to #490. It achieves ~93% of the encrypt savings with zero protocol changes. #490 achieves ~97% but requires a new wire format. Recommendation: ship #489 first (safe, incremental), then #490 as a protocol upgrade.
- **#486's commitment/RunningCommitments changes are independent** and always provide value regardless of which encrypt optimization lands.

## Ciphertext Format

```
ephemeralPubKey (65 bytes, uncompressed P-256 point)
|| txNonce     (16 bytes, random per-transaction for session key derivation)
|| iv          (12 bytes, AES-GCM nonce)
|| ciphertext  (AES-128-GCM encrypted payload + 16-byte auth tag)
```

`encryptionAlgorithmSpec` in the `AsymmetricEncrypted` protobuf message tells the receiver which decrypt path:
- `ECIES_HKDF_HMAC_SHA256_AES128CBC` (= 1): existing full ECIES
- `ECDH_CHANNEL_AES128GCM` (= 3): channel-key mode (this PR)

Old nodes reject channel-key ciphertexts cleanly (unrecognized protobuf enum). New nodes support both.

## Security

- **Per-transaction freshness**: random 16-byte txNonce ensures unique session keys
- **Forward secrecy**: `ChannelKeyStore.rotateEpoch()` on topology changes generates new ephemeral keys
- **Deterministic encryption**: not supported (channel keys are inherently non-deterministic)
- **Equivalent to**: TLS 1.3 session resumption, Signal pre-keys, MLS epoch keys

## Test plan

- [ ] Channel-key encrypt → decrypt round-trip produces correct plaintext
- [ ] Channel-key ciphertext rejected by old nodes (clean Unrecognized enum)
- [ ] Full ECIES ciphertext still works (no regression)
- [ ] ChannelKeyStore epoch rotation invalidates old channels

### DIFF
--- a/community/base/src/main/protobuf/com/digitalasset/canton/crypto/v30/crypto.proto
+++ b/community/base/src/main/protobuf/com/digitalasset/canton/crypto/v30/crypto.proto
@@ -290,6 +290,13 @@ enum EncryptionAlgorithmSpec {
   /* RSA with OAEP Padding,
      using SHA-256 for both the hash and in the MGF1 mask generation function along with an empty label. */
   ENCRYPTION_ALGORITHM_SPEC_RSA_OAEP_SHA256 = 2;
+
+  /* Channel-key mode: uses a cached per-participant-pair ECDH shared secret
+     (established on first contact) and derives per-transaction AES-128-GCM keys via HKDF.
+     The ciphertext contains: ephemeral_pub_key(65 bytes) || iv(12) || aes_gcm_ciphertext.
+     The receiver does ECDH with the embedded ephemeral key (or uses its cached channel key)
+     to derive the same AES key. ~72x faster than full ECIES after first contact. */
+  ENCRYPTION_ALGORITHM_SPEC_ECDH_CHANNEL_AES128GCM = 3;
 }
 
 // @deprecated
--- a/community/base/src/main/scala/com/digitalasset/canton/crypto/Encryption.scala
+++ b/community/base/src/main/scala/com/digitalasset/canton/crypto/Encryption.scala
@@ -373,6 +373,20 @@ object EncryptionAlgorithmSpec {
       v30.EncryptionAlgorithmSpec.ENCRYPTION_ALGORITHM_SPEC_RSA_OAEP_SHA256
   }
 
+  /** Channel-key mode: uses a cached ECDH shared secret per participant pair.
+    * After the first ECDH, subsequent encryptions derive per-transaction keys via HKDF,
+    * costing ~0.005ms instead of ~0.36ms per recipient.
+    * Ciphertext format: ephemeral_pub_key(65 bytes) || iv(12) || aes_gcm_ciphertext.
+    */
+  case object EcdhChannelAes128Gcm extends EncryptionAlgorithmSpec {
+    override val name: String = "ECDH_CHANNEL_AES128-GCM"
+    override val supportDeterministicEncryption: Boolean = false
+    override val supportedEncryptionKeySpecs: NonEmpty[Set[EncryptionKeySpec]] =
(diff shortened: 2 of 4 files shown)


elliottdehn:

elliottdehn:
## Impact Analysis: Single-Party High-Volume Scenario

For an exchange-like participant sending CC transfers through one party, 20 SVs, 2 views per transfer:

**Submitter encrypt cost per transaction:**

| Scenario | ECIES per tx | Other CPU | Total CPU/tx | 1-core tx/s |
|---|---|---|---|---|
| Today | 44 × 0.4ms = 17.6ms | ~2ms | ~20ms | ~50 |
| Channel keys, new counterparty | 2 × 0.4ms + 40 × 0.005ms = 1.0ms | ~2ms | ~3ms | ~330 |
| Channel keys, repeat counterparty | 44 × 0.005ms = 0.2ms | ~2ms | ~2.2ms | ~450 |

**On a 4-core machine:**

| Scenario | Throughput |
|---|---|
| Today | ~200 tx/s |
| Channel keys, new counterparties | ~1,300 tx/s |
| Channel keys, repeat counterparties | ~1,800 tx/s |

**6-9x throughput on the same hardware, no architectural changes.** The ECIES cost effectively vanishes for repeat counterparties, shifting the bottleneck from per-participant crypto to BFT consensus throughput.

For an exchange doing 10,000 CC sends/day with mostly repeat customers: ~84 seconds/day of ECIES today → ~5 seconds/day with channel keys.

elliottdehn:
Note: this PR adds a new `EncryptionAlgorithmSpec` protobuf enum value (`ECDH_CHANNEL_AES128GCM = 3` in `crypto.proto`). Per contributing guidelines, this needs review from `canton-change-owners`.

PR #501, opened 2026-04-16
Transparent gzip compression for sequencer payloads (#502)

### LINKED ISSUE #502: Daml-LF Value protobufs stored uncompressed in PostgreSQL
Serialized Daml-LF Value protobufs are stored as raw bytes in PostgreSQL. These values are structurally repetitive (same protobuf field tags, same template shapes across contracts) and compress well — 38-85% reduction with gzip depending on value size and structural similarity.

Smaller stored values → less DB IO → less network for replication → less memory for caches.

PR: #501 (draft)

elliottdehn:
Closes #502

## Summary

Transparent gzip compression for sequencer payloads stored in PostgreSQL. New writes are compressed. Old reads trigger lazy background compression. Zero-downtime migration — no operator action needed.

## How it works

**Flag byte protocol:**
- `0x00` + raw bytes = stored uncompressed (explicitly flagged)
- `0x01` + gzip bytes = stored compressed
- No flag (byte ≥ 0x08) = legacy uncompressed protobuf (backward compatible)

**Write path:** `ValueCompression.compress(payload.content)` before INSERT. Only compresses when it reduces size (values < 64 bytes stored raw).

**Read path:** `ValueCompression.decompressIfFlagged(content)` after SELECT. Legacy unflagged data passes through unchanged.

**Lazy migration:** When a read encounters legacy unflagged data, it queues the payload for background compression. A daemon thread drains the queue every 500ms, batching up to 100 UPDATEs per flush. Each payload compressed at most once — once the flag byte is written, subsequent reads skip the queue.

## Benchmark

| Payload type | Raw | Compressed | Savings | Decompress |
|---|---|---|---|---|
| Amulet (party+amount) | 106 B | 66 B | **38%** | 1.0µs |
| Medium (10 fields) | 263 B | 131 B | **50%** | 2.2µs |
| Large (50 fields) | 1343 B | 424 B | **68%** | 3.5µs |
| Batch (20 similar) | 2030 B | 313 B | **85%** | 3.0µs |

## Changes

| File | Change |
|---|---|
| `ValueCompression.scala` (new) | Flag byte compress/decompress, hasFlagByte, decompressIfFlagged |
| `DbSequencerStore.scala` | Write: compress before INSERT. Read: decompress after SELECT. Lazy migration queue + background flush thread. |

## Why it matters

Smaller payloads → less DB IO → less WAL → less replication traffic → more payloads fit in cache. At 1000 tx/s with 50% average compression, that's ~500KB/s less write throughput to PostgreSQL.

## Relation to other PRs

Independent of all crypto PRs (#486-493). This targets storage IO, not CPU. Stacks with everything.

## Test plan

- [x] Round-trip: compress → store → read → decompress produces original bytes
- [x] Legacy uncompressed data reads correctly (no flag byte, passes through)
- [x] Values < 64 bytes stored uncompressed (gzip overhead would expand them)
- [x] Lazy migration: legacy payload read → background UPDATE → next read has flag byte
- [x] `SimplestPingIntegrationTest` passes

### DIFF
--- a/community/base/src/main/scala/com/digitalasset/canton/util/ValueCompression.scala
+++ b/community/base/src/main/scala/com/digitalasset/canton/util/ValueCompression.scala
@@ -86,6 +86,36 @@ object ValueCompression {
     bytes.nonEmpty && bytes(0) != FLAG_UNCOMPRESSED
   }
 
+  /** Decompress if the data has a compression flag byte. If the first byte is
+    * not a recognized flag (i.e., it's a valid protobuf field tag), return the
+    * data as-is. This handles legacy uncompressed data transparently.
+    *
+    * Protobuf field tags are always >= 0x08 (field 1, wire type 0). Our flags
+    * are 0x00 (uncompressed) and 0x01 (gzip), which never appear as the first
+    * byte of valid protobuf.
+    */
+  def decompressIfFlagged(data: ByteString): ByteString = {
+    val bytes = data.toByteArray
+    if (bytes.isEmpty) return data
+
+    bytes(0) match {
+      case FLAG_UNCOMPRESSED =>
+        ByteString.copyFrom(bytes, 1, bytes.length - 1)
+      case FLAG_GZIP =>
+        try {
+          ByteString.copyFrom(gzipDecompress(bytes, 1, bytes.length - 1))
+        } catch {
+          case _: Exception =>
+            // If decompression fails, the data might be legacy uncompressed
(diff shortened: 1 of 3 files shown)


elliottdehn:

elliottdehn:
Closing — sequencer payloads are already gzip compressed (CompressedBatch) before storage. Compressing already-compressed data does nothing. The benchmark was against raw protobuf, not actual stored payloads.

Issue #513, opened 2026-04-30
Track external call support


angelol:
This issue tracked end-to-end support for external calls across the Canton/Daml stack, as the umbrella issue for TODOs and follow-up PRs while the feature was introduced in small reviewable steps. **The feature is now delivered end-to-end** (2026-07-07); the agreed post-merge work is tracked in #564 and #565.

## Completed upstream PRs

Foundations:
- #506 added the transaction-side representation for recorded external call results.
- #514 added the LF `EXTERNAL_CALL` builtin surface: proto/archive decoding, parser support, typing, spec documentation, and LF-focused tests.
- #518 added Engine and Speedy interpreter support. It wires the builtin into interpretation as a suspend/resume engine question and records successful external call results on the enclosing exercise node.
- #522 added transaction protobuf encoding and decoding support for recorded external call results.
- #526 added Canton protocol serialization support for recorded external call results using dev-only `ViewParticipantData` protocol data.
- #537 added participant-side extension-service client support for configured external call services.

Runtime integration (the former #541, superseded by a reviewable split #549–#556, completed by #567):
- #549 added transaction hashing scheme V4 and the `ViewParticipantData` wiring for recorded results.
- #550 added the prepared-transaction codec for external-call results.
- #551 added the command-execution errors, rejection routing and the external-call handler SPI.
- #552 added engine-side execution/replay, model-conformance integration and the validation SPI.
- #553 added the extension-service handler, configuration and command-execution wiring.
- #554 added the external-call consistency checker and shared validation fixtures.
- #567 completed the Canton-side runtime integration with the minimal protocol integration agreed with the reviewer: external-call validation runs as a parallel validation suite in transaction processing, disagreements reject the request on behalf of all hosted confirming parties, and results that cannot be re-validated lead to abstentions. #555 and #556 (per-party verdict routing) were closed as superseded by #567 — per-party granularity is deliberately dropped in favor of the fail-fast model tracked in #565.

Daml side:
- https://github.com/digital-asset/daml/pull/23099 added the SDK support: `DA.External.externalCall`, compiler/LF tooling support, and Daml Script submit-error reporting — merged against a Canton snapshot containing #567.

## Tracked follow-ups (post-merge, agreed during review)

- #564 — end-to-end and negative-path integration tests, plus the test cleanups from the #567 review.
- #565 — centralize disagreement validation in `TransactionView` (fail-fast) and the remaining review cleanup items.
[...]

PR #506, opened 2026-04-21
feat(external-call): Add transaction-side representation for external call results

angelol:
This PR is the transaction-side split of the earlier `external_call` work from [#438](https://github.com/digital-asset/canton/pull/438).

The broader feature here is support for deterministic external calls to extension services during transaction execution. Those results need to be recorded in the transaction so they can later be replayed during validation; this PR adds only that transaction-side representation.

It addresses the transaction-related review feedback from @paulbrauner-da on that PR by:
- keeping the transaction changes separate from the LF builtin/parser/typing changes
- using `bytes` rather than hex strings in the transaction format
- removing `call_index`, since protobuf repeated fields already preserve order

## What this PR does

- adds a transaction-side `ExternalCallResult` model on exercise nodes
- adds the corresponding transaction proto message
- stores `config`, `input`, and `output` as `bytes`
- keeps unsupported transaction-coder paths for external call results explicitly rejected for now

## Out of scope

This PR does not introduce the LF builtin or execution support for external calls.
Those are handled in follow-up PRs so this change remains small and reviewable on its own.

PR #514, opened 2026-04-30
Add LF builtin for external calls

angelol:
## Summary

This is the next preparatory layer for external calls after #506, which added the transaction-side representation for recording external call results.

This PR adds the Daml-LF representation of external calls by introducing a new `EXTERNAL_CALL` builtin. The builtin takes an extension identifier, function identifier, configuration hash, and input payload, all as `Text`, and returns the output payload in `Update Text`. The hash and payload arguments are represented as hexadecimal-encoded text.

The runtime execution path is intentionally not implemented in this PR to keep the scope focused. Until the engine/speedy support lands in a follow-up PR, the interpreter rejects `EXTERNAL_CALL` with an explicit compilation error.

## Changes

- Adds `EXTERNAL_CALL` to the LF v2 protobuf builtin enum, gated to LF 2.dev.
- Adds the corresponding LF AST builtin, parser support, typing rule, and archive decoding.
- Documents the builtin in the LF v2 spec.
- Extends LF archive/parser/typing/interpreter/encoder coverage and builtin fixtures.
- Updates the protobuf snapshot for the LF builtin addition.

Refs #513.


community/daml-lf/encoder/src/test/lf/Builtin_2.3-rc2_.lf
```diff
@@ -0,0 +1,138 @@
+// Copyright (c) 2025 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
```
remyhaemmerle-da: version  LF 2.3-rc2 will be gone soon. Please remove this file

community/daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/compiler/PhaseOneTest.scala
```diff
@@ -21,6 +21,25 @@ import scala.annotation.tailrec
...
+      new PhaseOne(interface, config)
+    }
+
+    val err = the[Compiler.CompilationError] thrownBy {
+      phase1.translateFromLF(PhaseOne.Env.Empty, EBuiltinFun(BExternalCall))
+    }
+
+    err.error shouldBe "EXTERNAL_CALL is not yet supported by the interpreter"
+  }
+
```
remyhaemmerle-da: We don't typically require tests for stubbed error cases. Feel free to handle these however you see fit.

community/daml-lf/validation/src/test/scala/com/digitalasset/daml/lf/validation/TypingSpec.scala
```diff
@@ -412,6 +412,15 @@ final class TypingSpec extends AnyWordSpec with TableDrivenPropertyChecks with M
       }
     }
 
+    "infer the proper type for EXTERNAL_CALL in 2.dev" in {
```
remyhaemmerle-da: we do test explicitly builtin. 
You just need to add an entry in `community/daml-lf/encoder/src/test/lf/Builtin_2.dev_.lf`

community/daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/compiler/PhaseOneTest.scala
angelol: Removed this stubbed error-case test from `PhaseOneTest`.

community/daml-lf/validation/src/test/scala/com/digitalasset/daml/lf/validation/TypingSpec.scala
angelol: Added the `EXTERNAL_CALL` coverage to `community/daml-lf/encoder/src/test/lf/Builtin_2.dev_.lf` and removed the bespoke `TypingSpec` case.

remyhaemmerle-da:
Thanks. 
Could you revert the changes in `TypingSpec.scala`  to be as it is in `main` ?

remyhaemmerle-da (dismissed):
LGTM

angelol:
Done in `d4d0c6c7e`: `TypingSpec.scala` is reverted back to match `main`, so it no longer appears in the PR diff.

angelol:
Rebased onto current `main` (`ee560db3f`) and pushed `b8e509a27`, which refreshes `.proto_snapshot_image.bin.gz` after the rebase.

remyhaemmerle-da (approved):

PR #521, opened 2026-05-06
Auto-detect latest GCP KMS cryptoKey version

same-id:
## Summary

Until now, every GCP KMS operation hardcoded cryptoKey version "1". This matches Canton-generated keys (Canton always creates a new cryptoKey rather than adding a new version to an existing one), but it prevents using cryptoKeys whose key material was imported into GCP KMS, since each import creates a new cryptoKey version and the version that holds the imported material is typically not "1" -- and can be relevant even for asymmetric keys.

Resolve the cryptoKey version dynamically via `listCryptoKeyVersions`, selecting the highest-numbered non-destroyed version of the cryptoKey. Once an ENABLED version is observed it is cached for the lifetime of the client, so signing and other hot-path operations only pay the list cost on first use. Symmetric encrypt/decrypt paths are unchanged: GCP picks the primary version on encrypt and recovers it from the ciphertext on decrypt, so no per-key version is needed there.

Adds the new `cloudkms.cryptoKeyVersions.list` IAM permission to the GCP KMS configuration reference and the external-key-storage how-to, plus an UNRELEASED.md entry.

### DIFF
--- a/community/base/src/main/scala/com/digitalasset/canton/crypto/kms/gcp/GcpKms.scala
+++ b/community/base/src/main/scala/com/digitalasset/canton/crypto/kms/gcp/GcpKms.scala
@@ -41,7 +41,9 @@ import org.bouncycastle.openssl.PEMParser
 import java.io.{IOException, StringReader}
 import java.util.UUID
 import scala.annotation.unused
+import scala.collection.concurrent.TrieMap
 import scala.concurrent.{ExecutionContext, Future, blocking}
+import scala.jdk.CollectionConverters.*
 
 /** Stands for Google Cloud Platform - Key Management Service and is an internal KMS implementation
   * that wraps the necessary cryptographic functions from the GCP SDK.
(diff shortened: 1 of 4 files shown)


community/base/src/main/scala/com/digitalasset/canton/crypto/kms/gcp/GcpKms.scala
```diff
@@ -73,10 +65,106 @@ class GcpKms(
...
+    * highest-numbered usable (i.e. non-destroyed) version of the cryptoKey. Once an `ENABLED`
+    * version is observed, it is cached for the lifetime of this client. Callers that need to
+    * react to state changes (e.g. `keyExistsAndIsActive`) should go through
+    * [[retrieveKeyMetadata]] which always re-issues the lookup.
+    */
+  private def resolveLatestKeyVersion(keyId: KmsKeyId)(implicit
+      ec: ExecutionContext,
+      tc: TraceContext,
+  ): EitherT[FutureUnlessShutdown, KmsError, String] =
+    resolvedKeyVersions.get(keyId.unwrap) match {
```
JoaoSa-da, line 105: I guess we could claim this is racy, but I think the key version here is effectively static, and we assume there won’t be changes to the version of that key in the future. So it’s fine if a put overwrites a previous value.

If you want to avoid unnecessary duplicate calls, we could try using `getOrElseUpdate` or something similar. But we can also leave it as it is, since I don’t expect much performance gain from optimizing this further.

community/base/src/main/scala/com/digitalasset/canton/crypto/kms/gcp/GcpKms.scala
```diff
@@ -73,10 +65,106 @@ class GcpKms(
...
+          keyId.unwrap,
+          keyMetadata.getAlgorithm.name,
+          keyMetadata.getState.name,
+        ),
+    )(
+      wrapKmsCall(
+        kmsErrorGen = (errStr, retryable) => KmsRetrieveKeyMetadataError(keyId, errStr, retryable),
+        functionName = functionFullName,
+      ) {
+        kmsClient.listCryptoKeyVersions(parent).iterateAll().asScala.toSeq
```
JoaoSa-da, line 143: You might need to rebase this because we just changed our GCP KMS client to use `kmsClient.xyzVersionCallable.futureCall`.

community/base/src/main/scala/com/digitalasset/canton/crypto/kms/gcp/GcpKms.scala
```diff
@@ -73,10 +65,106 @@ class GcpKms(
...
+        // Pick the highest version id (last segment of the name parses to an integer) so that
+        // both Canton-generated keys (always a single version "1") and imported keys (where
+        // each import adds a new version) work transparently. We sort numerically, not
+        // lexicographically, so version "10" comes after "9".
+        usable
+          .maxByOption(v => parseKeyVersionId(v.getName).toIntOption.getOrElse(-1))
+          .toRight[KmsError](
+            // Retryable so we don't hard-fail right after creation while the only version is
+            // still PENDING_GENERATION; the upstream retry loop will re-list and converge.
+            KmsRetrieveKeyMetadataError(
```
JoaoSa-da, line 159: I think that, if we want this to avoid becoming a breaking change, we need to have a fallback to version "1" in case we don’t have permission to use `listCryptoKeyVersions`.

community/base/src/main/scala/com/digitalasset/canton/crypto/kms/gcp/GcpKms.scala
```diff
@@ -610,125 +617,84 @@ class GcpKms(
...
+        )(
+          kmsClient.destroyCryptoKeyVersion(keyVersionName).discard
         )
-      ).map(_ => ())
-    )
-  }
+      )
+    }
 
+  /** Re-issues the list lookup (rather than going through the cached version) so that callers
```
JoaoSa-da, line 688: Not sure I understand why we need to make the call again instead of using the cache.

JoaoSa-da (commented):
It’s on the right track, but I don’t think we should call `listCryptoKeyVersions` whenever we can’t find the version in the cache. In my opinion, we should persist this version in the `KMSMetadataStore`, read it from the database, and then cache the result.

The `listCryptoKeyVersions` call should really only happen once during key import. Even for key auto-creation, we could potentially avoid calling it and instead default the version to "1".

I also think we should have a fallback to version "1" if we don’t have permission to call `listCryptoKeyVersions`, so we can guarantee backward compatibility.

same-id:
> It’s on the right track, but I don’t think we should call `listCryptoKeyVersions` whenever we can’t find the version in the cache. In my opinion, we should persist this version in the `KMSMetadataStore`, read it from the database, and then cache the result.

So it will not be possible to roll a new version of a key if we want to migrate because it will be forever stored in the canton DB? (from what I understand some keys can be rotated - basically everything except the namepsace key)

Not sure how you deal with rotation of keys such as the sequencer key, do both have to be registered in Canton node to do the rotation, how does that actually work.

> The `listCryptoKeyVersions` call should really only happen once during key import. Even for key auto-creation, we could potentially avoid calling it and instead default the version to "1".

Right
 
> I also think we should have a fallback to version "1" if we don’t have permission to call `listCryptoKeyVersions`, so we can guarantee backward compatibility.

I don't "love" this - just because if by some chance you have version 2, and version 1 is still enabled for some reason - you could have the node registering a version you didn't expect just because of permission issues.
I guess as a caller you can just disable version 1 before starting the canton node process, but it's not as explicit - just wanted you to see the diff between this approach which suddenly can fail on new api calls and is more implicit, and previous approach which is more explicit where we specify the version.

WDYT?

JoaoSa-da:
> So it will not be possible to roll a new version of a key if we want to migrate because it will be forever stored in the canton DB? (from what I understand some keys can be rotated - basically everything except the namepsace key)

Currently, key rotation is agnostic to the GCP crypto key version. For example, the Canton API for it:
```
def rotate_kms_node_key(
      fingerprint: String,
      newKmsKeyId: String,
      name: String = "",
  ): PublicKey
```
takes the fingerprint of the old key and replaces it by announcing, through topology management, the new key identified by the `newKmsKeyId` (i.e., add the new key with the KMS key ID to the topology, followed by removing the old one).

If we wanted to support GCP key versions, I guess we would need to change the behavior of this API so that we could specify the new key version, announce it as the new key, while locally updating the DB metadata entry to point to the new version.

I guess we don't need to support this right now. We can continue working with the assumption that rotating a key means creating a completely new key (with a different KMS ID), rather than creating a new version of the same KMS ID key. I think it would still be possible to change that in the future if it becomes a requirement.

> I don't "love" this - just because if by some chance you have version 2, and version 1 is still enabled for some reason - you could have the node registering a version you didn't expect just because of permission issues.
> I guess as a caller you can just disable version 1 before starting the canton node process, but it's not as explicit - just wanted you to see the diff between this approach which suddenly can fail on new api calls and is more implicit, and previous approach which is more explicit where we specify the version.

I agree with you that this risk exists, but AFAIK there is a strong constraint on our side against introducing breaking changes that could prevent some validator nodes from running. This would definitely be a breaking change because, after updating, a node might no longer be able to run if that permission is not granted. I can double-check, but I think we would prefer to live with that risk, which already exists today with the static "1" version, rather than introduce a breaking change.

same-id:
Ok,

Regarding:


> It’s on the right track, but I don’t think we should call listCryptoKeyVersions whenever we can’t find the version in the cache. In my opinion, we should persist this version in the KMSMetadataStore, read it from the database, and then cache the result.
>
> The listCryptoKeyVersions call should really only happen once during key import. Even for key auto-creation, we could potentially avoid calling it and instead default the version to "1".

What happens if I create a new key version? Once per startup will advance to the latest after listing. If I persist it, how do I update it?

JoaoSa-da:
> Ok,
> 
> Regarding:
> 
> > It’s on the right track, but I don’t think we should call listCryptoKeyVersions whenever we can’t find the version in the cache. In my opinion, we should persist this version in the KMSMetadataStore, read it from the database, and then cache the result.
> > The listCryptoKeyVersions call should really only happen once during key import. Even for key auto-creation, we could potentially avoid calling it and instead default the version to "1".
> 
> What happens if I create a new key version? Once per startup will advance to the latest after listing. If I persist it, how do I update it?

In what scenarios do you need to create a new key version? If you want to rotate a key, we currently assume you create a completely new key with a new KMS ID. The idea is that we always work with a single version per key (either version 1 if it is newly created, or another version if key material has been imported). If you want to rotate the key, you need to go through Canton; you cannot simply rotate it directly in the KMS, because you need to announce the new public keys to the network. In Canton we assume you need a new KMS ID and thus a totally new key.

JoaoSa-da:
Hey @same-id, just checking in on this PR - it’s been about a month since the last update. Are you still planning to work on it?

Happy to help unblock anything or answer any questions if needed. Let me know! :)

PR #522, opened 2026-05-11
feat(external-call): add transaction coding for external call results

angelol:
## Summary

This is the next small PR in the external-call stack tracked in #513.

#506 added the transaction-side representation for recording external call results. #514 added the LF `EXTERNAL_CALL` builtin surface. #518 wired the builtin into Speedy and the LF engine, including recording successful external call results on enclosing exercise nodes.

This PR adds transaction protobuf encoding and decoding support for those recorded external call results, so they survive transaction serialization before later Canton layers transmit, validate, and replay them.

## Scope

This PR adds:

- encoding of exercise-node `externalCallResults` in `TransactionCoder`
- decoding of exercise-node `externalCallResults` from transaction protobufs
- preservation of `extensionId`, `functionId`, `config`, `input`, and `output`
- serialization-version gating via `minExternalCallResults`
- focused transaction-coder tests for round-tripping and old-version rejection

## Out of scope

Canton protocol support, extension-service clients, runtime replay behavior, and validation policy are intentionally left for later PRs in the stack.

Refs #513.

PR #518, opened 2026-05-05
feat(external-call): add engine and Speedy support

angelol:
## Summary

This is the next small PR in the external-call stack tracked in #513.

#506 added the transaction-side representation for recording external call results. #514 added the LF `EXTERNAL_CALL` builtin surface. This PR wires that builtin into Speedy and the LF engine: interpretation can now suspend with a request for an external call result, then resume with either an output or an error.

Successful external call results are recorded on the nearest enclosing exercise node, so later stack layers can encode, transmit, validate, and replay them.

## Scope

This PR adds:

- Speedy support for `EXTERNAL_CALL`
- `NeedExternalCall` / `ResultNeedExternalCall` engine results
- strict canonical lowercase hex validation for config, input, and output payloads
- recording of successful external call results as transaction bytes
- serialization-version bumping for exercise nodes that contain external call results
- explicit rejection of external call results in V1 LF node hashing
- focused Speedy and engine tests for success, failure, validation, and exercise-context behavior

## Out of scope

The Canton participant/runtime integration is intentionally left for later PRs in the stack. Runtime execution, extension-service clients, replay behavior, and validation policy are not implemented here.

Current Canton runtime call sites remain fail-fast where external calls are not wired yet. Those paths are expected to be replaced by the later runtime integration PRs.

Refs #513.

PR #526, opened 2026-05-13
feat(external-call): add Canton protocol support for external call results

angelol:
## Summary

This is the next small PR in the external-call stack tracked in #513.

#506 added the transaction-side representation for recording external call results. #514 added the LF `EXTERNAL_CALL` builtin surface. #518 wired the builtin into Speedy and the LF engine. #522 added transaction protobuf encoding and decoding for recorded external call results.

This PR adds Canton protocol serialization support for those recorded external call results. After Andreas' review feedback, the protocol carrier is deliberately **not** `ActionDescription`: external-call records are stored in dev-only `v32.ViewParticipantData`, where the view core data belongs.

## Scope

This PR adds:

- a dev-only `v32.ViewParticipantData` protobuf version, following Canton's versioned-protobuf compatibility rules
- `ViewExternalCallResult` records containing call identity, `config`, `input`, `output`, view-local `node_id`, `call_index`, and node-level `checking_parties`
- Scala serialization/deserialization for the new `v32` carrier while leaving stable `v30` and `v31` protocol messages unchanged
- NextGen transaction-tree extraction of external-call results from the core nodes of each view
- validation guardrails for dev-only external-call records, occurrence identity, and exercise-root-only carrier use
- focused coverage for serialization, reconstruction, checking-party metadata, and the dev/stable protocol boundary
- the updated protobuf snapshot

## Out of scope

Extension-service clients, runtime replay behavior, external-call validation policy, and the remaining participant integration are intentionally left for later PRs in the stack.

Refs #513.


community/base/src/main/protobuf/com/digitalasset/canton/protocol/v30/participant_transaction.proto
```diff
@@ -179,6 +179,18 @@ message ActionDescription {
...
+
+  // Result of an external call recorded under an exercise node.
+  // Stores the call identity, config/input bytes, and output bytes for
+  // deterministic replay and validation.
+  message ExternalCallResult {
+    string extension_id = 1; // Extension service identifier
+    string function_id = 2; // Function identifier within the extension
+    bytes config = 3; // Extension configuration
+    bytes input = 4; // Input data sent to the extension
+    bytes output = 5; // Output data returned by the extension
```
andreaslochbihler-da: Unfortunately, it is almost always insecure to modify existing protobuf definitions of Canton protocol messages, even if the modification is considered backwards compatible in the Protobuf sense. The problem is that the Scala parsers generated by the protobuf tooling behave differently if they get garbage input, but consensus requires that they behave the same. This is explained in https://github.com/digital-asset/canton/blob/main/contributing/backwards-incompatible-changes.md#backwards-incompatible-protobuf-changes.

This means that you'll have to go through the trouble of creating a new proto version `v32` in this case, copy the relevant message definitions over, and wire up everything on the Scala side. https://github.com/digital-asset/canton/blob/main/contributing/backwards-incompatible-changes.md#protobuf-files-and-new-versions explains the tooling.

To give you an idea of what's needed, https://github.com/digital-asset/canton/pull/433/ has gone through the same exercise for `SubmitterMetadata`: I'd recommend to take a look at the changes to the files `community/base/src/main/scala/com/digitalasset/canton/data/SubmitterMetadata.scala` and `community/base/src/main/protobuf/com/digitalasset/canton/protocol/v31/participant_transaction.proto` in this PR.

community/base/src/main/scala/com/digitalasset/canton/data/ActionDescription.scala
```diff
@@ -342,6 +354,7 @@ object ActionDescription {
       override val byKey: Boolean,
       seed: LfHash,
       failed: Boolean,
+      externalCallResults: ImmArray[ExternalCallResult],
```
andreaslochbihler-da: `ActionDescription`s are not the appropriate place for storing the recorded external calls. Consider a transaction of the following shape:

* Root node `N1` exercises a choice on some contract. The node has two children `N2` and `N4`:
  * Node `N2` exercises another choice on some other contract. It records two external calls `EC1` and `EC2` and has one child:
    * Node `N3` exercises another choice on some other contract. It records two external calls `EC2` and `EC3. It does not have any children.
  * Node `N4` exercises yet another choice on some contract. It records one external call `EC4`. It does not have any children.

Suppose further that the informees of the nodes are as follows:
* `N1`: parties `A` and `B`
* `N2`: party `A`
* `N3`: parties `A` and `C`
* `N4`: party `B`

Let's consider the simple case where each party is hosted on their own node. Then Canton's view decomposition will likely create two views:
* `V1` with root `N1` and core `N1`, `N2`, `N4`.
* `V2` with root `N3` and core `N3`.

Now, `ActionDescription`s always refer to the root of a view only and contain the data that is needed to construct the relevant `ReplayCommand` for reinterpretation. So in the above example, we'd create two `ExerciseActionDescription`, one from `N1` and one from `N3`. So the recorded external calls results in `N2` and `N4` would be lost.

Instead, you probably want to record the external call results for all nodes in the core of a view in the `ViewParticipantData` in a separate field, because the `ViewParticipantData` stores the information collected from all nodes in the core of a view. So `V1` should store `EC1`, `EC2`, and `EC4`, and `V3` should store `EC2` and `EC3`.

Note that `EC2` appears in two nodes in this examples: `N2`, and `N3`, with different informees (and possibly confirmers). What is your strategy for validating the consistency of these calls? Depending on this strategy, you may want to annotate each external call result with the parties that are supposed to check the consistency.

Remark: This problem would go away if external call results propagated to parent exercises all the way up to the root, i.e., `N1` would include the recorded external call results of all its children, grand children, and so on. This is quite inefficient though, because `EC3` would be shipped twice as part of the protocol: once for view `V1` and once for view `V2`. I therefore do not consider this an option.

In fact, you should actually remove the external call result from a parent view if one of the (transitive) subviews already records the same result. In the above example, this means that `EC2` is only shipped as part of `V2`, but not as part of `V1` because anyone that receives `V1` also receives `V2` and may therefore extract `EC2` from `V2`.

andreaslochbihler-da (commented):
I'm afraid that this changeset needs to be redone:
* The protobuf definitions must be changed in a secure way.
* The external call results should be recorded in the view participant data, not in the action descriptions.

Please see my inline comments for the details.

community/base/src/main/protobuf/com/digitalasset/canton/protocol/v30/participant_transaction.proto
angelol: Thanks Andreas, addressed in the follow-up commit `70ebfed7f`.

I stopped modifying the existing `v30` protocol message and moved the new carrier to a dev-only `v32.ViewParticipantData` proto version, following the versioned-protobuf pattern from the `SubmitterMetadata` change you pointed to.

community/base/src/main/scala/com/digitalasset/canton/data/ActionDescription.scala
angelol: Thanks Andreas, addressed in the follow-up commit `70ebfed7f`.

External-call records are no longer stored in `ActionDescription`. They are now carried in dev-only `v32.ViewParticipantData` and collected from each view's core nodes with `node_id`, `call_index`, and node-level `checking_parties`, so validation can decide which hosted parties are responsible for each recorded call.

One deliberate adjustment: the factory preserves concrete occurrences per view instead of suppressing same-looking parent/subview records. Same-semantic-call disagreements are handled in validation, which avoids losing distinct LF occurrences with different checking parties.

community/base/src/main/protobuf/com/digitalasset/canton/protocol/v32/participant_transaction.proto
```diff
@@ -0,0 +1,47 @@
...
+
+// Compared to v31: external call results are recorded for the view core.
+message ViewParticipantData {
+  option (scalapb.message).companion_extends = "com.digitalasset.canton.version.AlphaProtoVersion";
+
+  message KeyResolutionWithMaintainers {
+    v31.GlobalKey key = 1;
+    repeated string maintainers = 2;
+    repeated string contract_ids = 3;
+  }
```
andreaslochbihler-da: Minor: Since this message is not modified, you could avoid copying it and reuse the one from `v31`. Then you wouldn't have to duplicate the `from/toProtoV32` methods in `KeyResolutionWithMaintainers`.

community/base/src/main/scala/com/digitalasset/canton/data/ViewParticipantData.scala
```diff
@@ -128,6 +134,28 @@ final case class ViewParticipantData private (
         s"Contract created in a subview are also created in the core: $transientOverlap"
       )
 
+    if (
+      externalCallResults.nonEmpty &&
+      representativeProtocolVersion.representative != ProtocolVersion.dev
```
andreaslochbihler-da: It's always dangerous to compare a representative protocol version's representative to a fixed protocol version `pv`. This works well as long as the equivalence class of protocol versions contains just `pv`, but this normally changes over time. You want to convert `pv` first to a representative as shown for example in https://github.com/digital-asset/canton/blob/8d252e6592354ac244eb65c850c4ef2a9a9b63c2/community/base/src/main/scala/com/digitalasset/canton/data/ViewParticipantData.scala#L155-L161. Moreover `==` is also bad because we're going to introduce new protocol versions and then we don't want to modify all the checks for previous protocol version.


```suggestion
      representativeProtocolVersion < ViewParticipantData.representativeProtocolVersion(ProtocolVersion.dev)
```

community/base/src/main/scala/com/digitalasset/canton/data/ViewParticipantData.scala
```diff
@@ -128,6 +134,28 @@ final case class ViewParticipantData private (
...
+    )
+      throw InvalidViewParticipantData(
+        s"External call results are supported only in protocol version ${ProtocolVersion.dev}"
```
andreaslochbihler-da:
```suggestion
        s"External call results are supported only from protocol version ${ProtocolVersion.dev} onwards"
```

community/base/src/main/scala/com/digitalasset/canton/data/ViewParticipantData.scala
```diff
@@ -576,6 +672,74 @@ object ViewParticipantData
   /** Indicates an attempt to create an invalid [[ViewParticipantData]]. */
   final case class InvalidViewParticipantData(message: String) extends RuntimeException(message)
 
+  final case class ViewExternalCallResult(
+      result: ExternalCallResult,
+      nodeId: LfNodeId,
+      callIndex: Int,
```
andreaslochbihler-da: Since you throw on negative call indices, did you consider to use `NonNegativeInt` as a type? Then you don't need to explicitly check this as an invariant in this class.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
```diff
@@ -378,6 +424,18 @@ class NextGenTransactionTreeFactory(
               case lfNode: LfActionNode =>
                 val suffixedNode = trySuffixNode(state)(nodeId -> lfNode)
                 coreOtherBuilder += ((nodeId, lfNode) -> rbScope)
+                if (collectExternalCallResults)
+                  lfNode match {
+                    case exercise: LfNodeExercises if exercise.externalCallResults.nonEmpty =>
+                      val submittingAdminPartyForNodeO =
+                        if (sourceRootNodeIds(nodeId)) submittingAdminPartyO else None
```
andreaslochbihler-da: I do not really understand why the submitting admin party is threaded through here. The submitting admin party is merely added to some views so that the submitting participant can veto the confirmation request after sequencing. It has nothing to do with the authorization logic encoded in the Daml model and should therefore not influence the checking parties for external calls.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
```diff
@@ -820,6 +887,111 @@ class NextGenTransactionTreeFactory(
...
+  private[submission] def nodeIdNormalizationForView(
+      transaction: LfVersionedTransaction,
+      viewRootNodeId: LfNodeId,
+      nodeIds: Set[LfNodeId],
+  ): Map[LfNodeId, LfNodeId] =
+    if (nodeIds.isEmpty) Map.empty
+    else {
+      val normalizedNodeIds = Map.newBuilder[LfNodeId, LfNodeId]
+      var nextNodeIndex = 0
+      var remainingNodeIds = nodeIds
```
andreaslochbihler-da: Stylistic: These two `var`s are essentially the loop state of `go`. Consider to move them to arguments of `go` so that you don't need `var`s with the wart supression.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
```diff
@@ -297,6 +308,22 @@ class NextGenTransactionTreeFactory(
 
     // contract IDs have not yet been suffixed
     val coreOtherBuilder = List.newBuilder[((LfNodeId, LfActionNode), RollbackScope)]
+    val collectExternalCallResults =
+      protocolVersion.isDev && sourceTransaction.version == LfSerializationVersion.VDev
```
andreaslochbihler-da:
```suggestion
      protocolVersion >= ProtocolVersion.dev && sourceTransaction.version >= LfSerializationVersion.VDev
```

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
```diff
@@ -297,6 +308,22 @@ class NextGenTransactionTreeFactory(
...
+
+    def coreExternalCallResultsBuilder()
+        : mutable.Builder[CoreExternalCallResult, List[CoreExternalCallResult]] =
+      coreExternalCallResultsBuilderO match {
+        case Some(builder) => builder
+        case None =>
+          val builder = List.newBuilder[CoreExternalCallResult]
+          coreExternalCallResultsBuilderO = Some(builder)
+          builder
+      }
```
andreaslochbihler-da: This construction looks quite complicated. Are you sure that it is worth the effort to go through the `Option`s and checks just to save on creating a `List.newBuilder` when there are no external calls? 

Did you consider using a `lazy val coreExternalCallResultsBuilder = List.newBuilder[CoreExternalCallResult]` to achieve the same with a lot less boilerplate?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
```diff
@@ -287,6 +296,8 @@ class NextGenTransactionTreeFactory(
       state: State,
       contractOfId: ContractInstanceOfId,
       topologySnapshot: TopologySnapshot,
+      sourceTransaction: LfVersionedTransaction,
+      submittingAdminPartyO: Option[LfPartyId],
```
andreaslochbihler-da: I'm not particularly keen on seeing the `sourceTransaction` being passed in here in addition to `view`, because the two parameters now contain redundant information: the `TransactionViewDecomposition` encodes the transaction and `sourceTransaction` does so too. This duplication creates overhead in tests (where one has to make sure that the decomposition fits to the source transaction) and potential for programming mistakes where discrepancies between the view decomposition and the source transaction can lead to subtle bugs.

It would be better to extend the `TransactionViewDecomposition` with the relevant information needed for collecting the external call results. AFAICT the `sourceTransaction` is only needed to normalize the node IDs in the view. If you use an "exercise index" akin to the [`createIndex` for create nodes](https://github.com/digital-asset/canton/blob/8d252e6592354ac244eb65c850c4ef2a9a9b63c2/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala#L324) in the core of a view, you wouldn't need this parameter and the whole node ID normalization.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
```diff
@@ -820,6 +887,111 @@ class NextGenTransactionTreeFactory(
...
+      val result = normalizedNodeIds.result()
+      val missingNodeIds = nodeIds -- result.keySet
+      if (missingNodeIds.nonEmpty)
+        throw new IllegalStateException(
+          s"Did not find external call node ids in view-local node map: $missingNodeIds"
+        )
+      result
+    }
+
+  private[submission] final case class CoreExternalCallResult(
```
andreaslochbihler-da: What is the difference to `ViewParticipantData.ViewExternalCallResult`? Can the two case classes be unified?

community/common/src/test/scala/com/digitalasset/canton/data/TransactionViewTest.scala
```diff
@@ -164,6 +214,8 @@ class TransactionViewTest
         createdIds: Seq[LfContractId] = Seq(createdId),
         archivedInSubviews: Set[LfContractId] = Set.empty,
         resolvedKeys: Map[LfGlobalKey, LfVersioned[KeyResolutionWithMaintainers]] = Map.empty,
+        externalCallResults: ImmArray[ViewParticipantData.ViewExternalCallResult] = ImmArray.Empty,
+        protocolVersion: ProtocolVersion = testedProtocolVersion,
```
andreaslochbihler-da: We try to not hard-code protocol versions in tests, but always use `testedProtocolVersion`. This is because our CI runs tests with all protocol versions of interest by setting `testedProtocolVersion` via environment variables. If you want to restrict certain tests to certain protocol versions, use `onlyRunWithOrGreaterThan ProtocolVersion.dev` as can be seen e.g. in https://github.com/digital-asset/canton/blob/8d252e6592354ac244eb65c850c4ef2a9a9b63c2/community/common/src/test/scala/com/digitalasset/canton/data/TransactionViewTest.scala#L299.

community/common/src/test/scala/com/digitalasset/canton/data/TransactionViewTest.scala
```diff
@@ -327,6 +420,143 @@ class TransactionViewTest
...
+        forEvery(Seq(ProtocolVersion.v34, ProtocolVersion.v35)) { protocolVersion =>
+          val vpd = create(protocolVersion = protocolVersion).value
+
+          ViewParticipantData
+            .fromByteString(protocolVersion, hashOps)(
+              vpd.getCryptographicEvidence
+            )
+            .map(_.unwrap.map(_.externalCallResults)) shouldBe Right(Right(ImmArray.Empty))
+        }
+      }
```
andreaslochbihler-da: This test seems redundant, as our CI will run this test for each protocol version by setting `testedProtocolVersion`.

community/base/src/main/protobuf/com/digitalasset/canton/protocol/v32/participant_transaction.proto
```diff
@@ -0,0 +1,47 @@
...
+message ViewExternalCallResult {
+  option (scalapb.message).companion_extends = "com.digitalasset.canton.version.AlphaProtoVersion";
+
+  string extension_id = 1;
+  string function_id = 2;
+  bytes config = 3;
+  bytes input = 4;
+  bytes output = 5;
+  optional int32 node_id = 6;
+  optional int32 call_index = 7;
```
andreaslochbihler-da: Why are these two marked as `optional`? The test cases seem to indicate that they must always be set.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactoryTest.scala
```diff
@@ -108,11 +185,413 @@ final class NextGenTransactionTreeFactoryTest
...
+            val treeFactory = createTransactionTreeFactory(devFactory)
+            val example = devFactory.MultipleRootsAndSimpleViewNesting
+            val nodeId = LfNodeId(5)
+
+            createTransactionTree(
+              treeFactory,
+              withExternalCallResults(
+                example,
+                nodeId,
+                ImmArray(externalCallResult, externalCallResult),
```
andreaslochbihler-da, line 245: Higher-level question: What is the point of recording the same `externalCallResult` multiple times for the same `nodeId`? This seems to be pretty redundant as even the checking parties are the same in this case.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ExampleTransactionConformanceTest.scala
```diff
@@ -190,15 +272,21 @@ class ExampleTransactionConformanceTest
           commonData,
           getEngineAbortStatus = () => EngineAbortStatus.notAborted,
           reInterpretedTopLevelViews,
-          testedProtocolVersion,
+          protocolVersion,
```
andreaslochbihler-da: Same comment as above: Please stick to `testedProtocolVersion` and control the tests with `onlyRunWithOrGreaterThan` and friends.

community/base/src/main/scala/com/digitalasset/canton/data/ViewParticipantData.scala
```diff
@@ -399,6 +448,7 @@ object ViewParticipantData
       rollbackContext: RollbackContext,
       salt: Salt,
       protocolVersion: ProtocolVersion,
+      externalCallResults: ImmArray[ViewExternalCallResult] = ImmArray.Empty,
```
andreaslochbihler-da: Can you avoid default arguments for internal methods? They make it harder to review changes because not all call site show up in a change.

For example, this default argument hides that this PR should also extend the scalacheck generator for `ViewParticipantData` in https://github.com/digital-asset/canton/blob/8d252e6592354ac244eb65c850c4ef2a9a9b63c2/community/common/src/test/scala/com/digitalasset/canton/data/GeneratorsData.scala#L248.

community/base/src/main/scala/com/digitalasset/canton/data/ViewParticipantData.scala
```diff
@@ -436,6 +488,7 @@ object ViewParticipantData
...
```
andreaslochbihler-da: Same as above: Please avoid default arguments in methods.

andreaslochbihler-da (commented):
This looks much better than the previous version. Thanks a lot for addressing the problems. At a high level, I've made the following observations:

1. The submitting admin party seems to influence the checking parties of external call results recorded for root nodes. I did not understand the rationale for this. Since external calls happen only in the body of the exercised choice, the normal Daml authorization rules should apply and therefore the checking parties should be the signatories of the input contract and the actors of the choice, shouldn't they? Unless I'm missing an important point, I have the impression that removing this special case can simplify the implementation quite a bit.
2. It looks suboptimal to me to use node IDs to refer to the external call results of different exercise nodes in the core of a view. This leads to redundancy in the function arguments and complicates the code because node IDs must be normalized (which seems to result in an algorithm that's quadratic in the depth of the tree). From what I can see, an exercise index that simply counts the exercise nodes in the core of a view would be a much simpler solution.
3. There is no attempt to avoid storing identical external call results if the same external call is made multiple times in a core of a view or between subview and parent view (see https://github.com/digital-asset/canton/pull/526#discussion_r3297666926). Within the core of a view, I don't see a big problem with this because the encoded `ViewParticipantData` is compressed before encryption and compression is good at removing such redundancies. Yet, no such compression happens between parent and subview. Given the discussion on slack that the amount of data passed over the external call interface can be substantial, this can lead to substantial increases in traffic costs for the submitters. However, I do not yet know practical use cases for external call and therefore cannot judge whether this potential waste of traffic credits actually matters in practice.

I've additionally left a few further comments on organization and testing.

community/base/src/main/protobuf/com/digitalasset/canton/protocol/v32/participant_transaction.proto
angelol: `v32.ViewParticipantData` now reuses `v31.ViewParticipantData.KeyResolutionWithMaintainers`, and the duplicate `KeyResolutionWithMaintainers.toProtoV32` / `fromProtoV32` helpers were removed.

community/base/src/main/scala/com/digitalasset/canton/data/ViewParticipantData.scala
angelol: The check now compares against the representative protocol version for ProtocolVersion.dev and uses a range check, so future protocol versions in later representatives do not require revisiting this condition.

community/base/src/main/scala/com/digitalasset/canton/data/ViewParticipantData.scala
angelol: I applied the suggested wording: external call results are now reported as supported only from ProtocolVersion.dev onwards.

community/base/src/main/scala/com/digitalasset/canton/data/ViewParticipantData.scala
angelol: ViewExternalCallResult now stores both exerciseIndex and callIndex as NonNegativeInt, and proto decoding uses parseNonNegativeInt, so the manual negative-value invariant checks are gone.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
angelol: The submitting admin party is no longer included in external-call checking parties. Checking parties are now derived only from the LF exercise node’s signatories/maintainers and acting parties.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
angelol: The node-id normalization helper containing this loop state was removed. The factory now records a view-local exerciseIndex directly while traversing the view core, so these vars and the associated wart suppression are gone.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
angelol: The collection path now keeps the protocol-version guard and checks the individual exercise node’s LF serialization version before reading external-call results.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
angelol: Thanks. The Option-based builder plumbing was removed. The code now uses a simple lazy List.newBuilder[ViewParticipantData.ViewExternalCallResult].

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
angelol: sourceTransaction is no longer threaded through this path. The factory now records a view-local exerciseIndex while traversing the view core, analogous to createIndex, so the raw node-id normalization helper and the redundant transaction input were removed.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
angelol: CoreExternalCallResult was removed. The factory now constructs ViewParticipantData.ViewExternalCallResult directly while traversing the view core.

community/common/src/test/scala/com/digitalasset/canton/data/TransactionViewTest.scala
angelol: The dev-only external-call tests now use testedProtocolVersion and are guarded with onlyRunWithOrGreaterThan ProtocolVersion.dev.

community/common/src/test/scala/com/digitalasset/canton/data/TransactionViewTest.scala
angelol: The redundant explicit protocol-version loop was removed. The test suite now relies on Canton’s protocol-version test matrix via testedProtocolVersion.

community/base/src/main/protobuf/com/digitalasset/canton/protocol/v32/participant_transaction.proto
angelol: These fields are no longer optional. The proto now uses required-by-convention int32 exercise_index and int32 call_index, with Scala decoding into NonNegativeInt.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactoryTest.scala
angelol: Thanks, good question. The repeated payload is redundant at the byte level, but it represents two concrete LF external-call occurrences: LF stores external-call results as an ordered list on the exercise node.

I changed the protocol metadata to make that explicit. We no longer record repeated raw nodeIds; each record is identified by (exerciseIndex, callIndex), where callIndex is the position in that exercise node’s external-call result list. Exact duplicate occurrence IDs are rejected.

For this same-view/same-node case, the remaining byte-level redundancy is within one ViewParticipantData, so it should benefit from the compression you mentioned in the review summary. I therefore kept construction occurrence-preserving and avoided payload-based suppression in the factory; runtime validation/replay can still group by semantic call identity where needed.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ExampleTransactionConformanceTest.scala
angelol: The conformance regression now uses testedProtocolVersion and gates the external-call case with onlyRunWithOrGreaterThan ProtocolVersion.dev.

community/base/src/main/scala/com/digitalasset/canton/data/ViewParticipantData.scala
angelol: The default externalCallResults = ImmArray.Empty argument was removed from tryCreate, and call sites now pass the value explicitly, including the ScalaCheck generator.

community/base/src/main/scala/com/digitalasset/canton/data/ViewParticipantData.scala
angelol: Same fix here: the default argument was removed from create, so call sites now explicitly show whether they pass ImmArray.Empty or actual external-call records.

community/base/src/main/scala/com/digitalasset/canton/data/ViewParticipantData.scala
andreaslochbihler-da, line 377: Do you want to add another line here to pretty-print the external call result?

community/common/src/test/scala/com/digitalasset/canton/data/GeneratorsData.scala
```diff
@@ -245,6 +246,43 @@ final class GeneratorsData(
...
+              input = Bytes.fromStringUtf8(input),
+              output = Bytes.fromStringUtf8(output),
+            ),
+            checkingParties,
+          )
+        }.map(results =>
+          ImmArray.from(results.zipWithIndex.map { case ((result, checkingParties), callIndex) =>
+            ViewParticipantData.ViewExternalCallResult(
+              result = result,
+              exerciseIndex = NonNegativeInt.zero,
```
andreaslochbihler-da: Nit: Do you also want to vary the exercise index? These generators are primarily used to test serialization and deserialization and picking `zero` is particularly bad because that's the default value of a missing Protobuf field. So these test cases would not be able to catch if the serialization function forgot to set the `exerciseIndex`. Already choosing a non-zero value would be better than `zero`.

community/common/src/test/scala/com/digitalasset/canton/data/TransactionViewTest.scala
```diff
@@ -275,6 +334,46 @@ class TransactionViewTest
...
+    }
+
+    "external call results on a non-dev protocol version" must {
+      "reject creation" in {
+        create(
+          actionDescription = exerciseActionDescription,
+          coreInputs = exerciseCoreInputs,
+          externalCallResults = ImmArray(viewExternalCallResult(exerciseIndex = 7)),
+          protocolVersion = ProtocolVersion.v35,
```
andreaslochbihler-da: Similar to https://github.com/digital-asset/canton/pull/526#discussion_r3304619692: It would be better to condition this test with `onlyRunWithOrLessThan ProtocolVersion.v35` than to hard-code a protocol version, so that this works automagically by Canton's protocol version test matrix.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
```diff
@@ -725,12 +758,15 @@ class NextGenTransactionTreeFactory(
       viewSalts,
     )
 
+    val submittingAdminPartyO =
+      submittingAdminPartyForReconstruction(submittingParticipantO, rootPosition)
```
andreaslochbihler-da: Why is this change needed here? Is this a renmant from an earlier version where the submitting admin party was taken into account?

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ExampleTransactionConformanceTest.scala
```diff
@@ -178,6 +253,15 @@ class ExampleTransactionConformanceTest
...
         reInterpretedTopLevelViews: ModelConformanceChecker.LazyAsyncReInterpretationMap = Map.empty,
+    ): EitherT[Future, ErrorWithSubTransaction[Unit], Result] =
+      checkWithProtocolVersion(mcc, views, ips, reInterpretedTopLevelViews, testedProtocolVersion)
+
+    def checkWithProtocolVersion(
+        mcc: ModelConformanceChecker,
+        views: NonEmpty[Seq[(FullTransactionViewTree, Seq[(TransactionView, LfGlobalKeyMapping)])]],
+        ips: TopologySnapshot,
+        reInterpretedTopLevelViews: ModelConformanceChecker.LazyAsyncReInterpretationMap,
+        protocolVersion: ProtocolVersion,
```
andreaslochbihler-da: This method seems to be called only with `protocolVersion = testedProtocolVersion`. So consider to unify it again with `check`.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ExampleTransactionConformanceTest.scala
```diff
@@ -303,6 +393,96 @@ class ExampleTransactionConformanceTest
...
+          cantonContractIdVersion = contractIdVersion,
+        )
+        val devTreeFactory = TransactionTreeFactory(
+          ExampleTransactionFactory.submittingParticipant,
+          devFactory.psid,
+          devFactory.cantonContractIdVersion,
+          devFactory.cryptoOps,
+          TestContractHasher.Async,
+          loggerFactory,
+        )
```
andreaslochbihler-da: This factory looks identical to the `transactionTreeFactory` defined above. Can you reuse the global one and remove the unnecessary changes coming from this in this file?

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactoryTest.scala
```diff
@@ -108,11 +186,433 @@ final class NextGenTransactionTreeFactoryTest
...
           }
+
+          "record external call results from same-view exercise nodes with view-local occurrence indexes" onlyRunWithOrGreaterThan ProtocolVersion.dev in {
+            val devFactory = new ExampleTransactionFactory(
+              versionOverride = Some(testedProtocolVersion)
+            )(
+              psid = factory.psid.copy(protocolVersion = testedProtocolVersion),
+              cantonContractIdVersion = contractIdVersion,
+            )
+            val treeFactory = createTransactionTreeFactory(devFactory)
```
andreaslochbihler-da: Same as for `ExampleTransactionConformanceTest`: With the changes around `testedProtocolVersion`, this factory seems to be the same as `factory` and so the generalization of the methods above is actually not needed. Can you undo them again?

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactoryTest.scala
```diff
@@ -108,11 +186,433 @@ final class NextGenTransactionTreeFactoryTest
...
+            )
+            val example =
+              devFactory.SingleExerciseWithNonstakeholderActor(devFactory.deriveNodeSeed(0))
+            val transaction = withExternalCallResults(
+              example.versionedUnsuffixedTransaction,
+              example.metadata,
+              LfNodeId(0),
+              ImmArray(externalCallResult),
+              _.copy(actingParties = Set(ExampleTransactionFactory.signatory)),
+            )
```
andreaslochbihler-da: What's the point of modifying the acting parties here? The confirming parties of the root node here are the submitter and the signatory, so this test case seems to create an impossible case where the checking parties are wrong.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactoryTest.scala
```diff
@@ -108,11 +186,433 @@ final class NextGenTransactionTreeFactoryTest
...
+                  val record =
+                    reconstructedView.viewParticipantData.tryUnwrap.externalCallResults.toSeq.loneElement
+
+                  reconstructedView shouldBe submittedView
+                  record.exerciseIndex shouldBe NonNegativeInt.tryCreate(1)
+                }
+            }
+          }
+
+          "reconstruct nested view root external call records without the submitting participant admin party" onlyRunWithOrGreaterThan ProtocolVersion.dev in {
```
andreaslochbihler-da: Is this test case still needed now that the special treatment of the submitting participant admin party has been dropped?

andreaslochbihler-da (commented):
Thanks a lot for making all these changes. I've added a few more comments on how to polish the changes a bit.

We've also kicked off a CI run and quite a few jobs are red. I suspect that most of the test failures are due to the problems with the version of `main` that this branch has been forked off. The exception is the `scalafix` job, which you should be easily able to fix yourself by running `sbt format`. Therefore, I'd suggest that you incorporate the latest changes to `main` into this branch in addition to the suggestions below and then I can trigger another CI run. Does this work for you?

community/base/src/main/scala/com/digitalasset/canton/data/ViewParticipantData.scala
angelol: Added a compact pretty-printer entry for external call results, showing the extension/function and view-local occurrence as `extension:function@exercise.call`.

community/common/src/test/scala/com/digitalasset/canton/data/GeneratorsData.scala
angelol: Changed the generator to choose a non-zero `exerciseIndex` (`1` to `10`), so the serialization round-trip can catch missing/defaulted protobuf fields.

community/common/src/test/scala/com/digitalasset/canton/data/TransactionViewTest.scala
angelol: Changed this to use `onlyRunWithOrLessThan ProtocolVersion.v35` with `testedProtocolVersion`. I also added the missing helper to `ProtocolVersionChecksAnyWordSpec`, matching the other protocol-version test traits.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
angelol: Yes, this was a leftover from the earlier submitting-admin-party handling. Removed the extra change; reconstruction now uses the existing submitting participant admin-party argument shape again.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ExampleTransactionConformanceTest.scala
angelol: Unified this back into `check`; it now passes `testedProtocolVersion` directly and the separate helper is gone.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ExampleTransactionConformanceTest.scala
angelol: Reused the shared `transactionTreeFactory` here and removed the local duplicate factory.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactoryTest.scala
angelol: Undone. The test helper now uses the local `factory` directly again, and the unnecessary factory/participant generalization is gone.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactoryTest.scala
angelol: Removed that impossible test case. The remaining signatory/actor coverage now uses the existing `SingleExerciseWithNonstakeholderActor` fixture instead of mutating `actingParties` by hand.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactoryTest.scala
angelol: Removed this test case as well; it was specific to the dropped submitting-admin-party special handling.

angelol:
> Thanks a lot for making all these changes. I've added a few more comments on how to polish the changes a bit.
> 
> We've also kicked off a CI run and quite a few jobs are red. I suspect that most of the test failures are due to the problems with the version of `main` that this branch has been forked off. The exception is the `scalafix` job, which you should be easily able to fix yourself by running `sbt format`. Therefore, I'd suggest that you incorporate the latest changes to `main` into this branch in addition to the suggestions below and then I can trigger another CI run. Does this work for you?

Thanks for the review and the detailed pointers.

I rebased the branch onto current `main`, ran `sbt format`, and addressed the inline polish comments in the follow-up commit.

The branch is force-pushed now, so it should be ready for another CI run.

community/base/src/main/scala/com/digitalasset/canton/data/ViewParticipantData.scala
andreaslochbihler-da: The canonical way of pretty printing would be to define a `Pretty` instance for `ExternalCallResult` and just write `paramIfNonEmpty("external call result", _.externalCallResults.toSeq)`, but the current implementation also makes sense as it's quite concise. So let's leave it as is.

community/common/src/test/scala/com/digitalasset/canton/data/TransactionViewTest.scala
```diff
@@ -164,6 +219,8 @@ class TransactionViewTest
         createdIds: Seq[LfContractId] = Seq(createdId),
         archivedInSubviews: Set[LfContractId] = Set.empty,
         resolvedKeys: Map[LfGlobalKey, LfVersioned[KeyResolutionWithMaintainers]] = Map.empty,
+        externalCallResults: ImmArray[ViewParticipantData.ViewExternalCallResult] = ImmArray.Empty,
+        protocolVersion: ProtocolVersion = testedProtocolVersion,
```
andreaslochbihler-da: AFAICS all calls to `create` set `protocolVersion` to `testedProtocolVersion`, so this parameter seems redundant. Please consider to remove it.

andreaslochbihler-da (commented):
Thanks @angelol for polishing this PR. I've just one tiny clean-up comment. Can you quickly address it and then we should be ready to merge this PR.

community/common/src/test/scala/com/digitalasset/canton/data/TransactionViewTest.scala
angelol: Removed the redundant parameter; the local `create` helper now uses `testedProtocolVersion` directly, and the call sites no longer pass it explicitly.

andreaslochbihler-da (approved):

PR #541, opened 2026-06-11
feat(external-call): add runtime integration

angelol:
## Summary

This is the concluding Canton-side PR in the external-call stack tracked in #513, stacked on #537.

#506 added the transaction-side representation for recording external call results. #514 added the LF `EXTERNAL_CALL` builtin surface. #518 wired the builtin into Speedy and the LF engine. #522 added transaction protobuf encoding and decoding for recorded external call results. #526 added Canton protocol serialization support for those recorded results. #537 adds participant-side extension-service client support.

This PR completes the Canton-side implementation by wiring the feature into participant runtime execution. Participants can now route engine external-call questions to configured extension services, record and replay results through prepared transactions, and validate recorded outputs during confirmation.

## Scope

This PR adds:

- participant extension-service configuration keyed by extension id
- HTTP extension-service clients with per-extension connection management, TLS trust configuration, bearer-token-file auth, timeouts, retries, and optional startup validation
- runtime handling for engine external-call questions in command interpretation
- external-call result encoding/decoding in prepared transactions and interactive submission data
- external-call-aware command execution errors and error-resource metadata
- replay and validation plumbing for recorded external-call results
- consistency checking that compares recorded results with locally checked results and routes disagreements to the relevant checking parties
- node-hashing and transaction-hashing integration for external-call records
- participant and ledger-api-core tests for configuration validation, HTTP behavior, prepared transaction codec behavior, runtime dispatch, conformance, and validation disagreement handling
- the updated protobuf snapshot

## Hashing Boundary

Prepared/external-submission signatures authorize the LF action. Recorded external-call result payloads are Canton protocol data: after submission they are carried through `ViewParticipantData.externalCallResults` and protected by the Canton view signature. For that reason prepared transactions can carry external-call results for round-tripping, but those result payloads are intentionally excluded from the LF/prepared node hash.

## Review Aid

For reviewers who want to inspect only the incremental runtime-integration diff on top of #537:

- GitHub compare view under the upstream repository namespace: https://github.com/digital-asset/canton/compare/zenith-network:angelol/external-call-05-extension-service-redesign...zenith-network:angelol/external-call-06-runtime-integration-redesign
- Review-only draft PR in the fork, with normal PR review UI: https://github.com/zenith-network/canton/pull/17

The review-only PR compares this branch against `angelol/external-call-05-extension-service-redesign` and is not intended to be merged.

## Coordination

This PR must merge before the companion Daml PR can merge: https://github.com/digital-asset/daml/pull/23099.

This PR is stacked on #537 and should be rebased after #537 merges. The Daml PR can remain open for parallel review, but after this Canton PR lands it must be amended to pin a normal Canton snapshot/artifact containing these runtime constants and error contracts. Daml verification should then be rerun against that published artifact before the Daml PR is merged.

Refs #513.


community/participant/src/main/scala/com/digitalasset/canton/participant/extension/HttpExtensionServiceClient.scala
```diff
@@ -0,0 +1,527 @@
...
+        )
+        ExtensionCallError(500, "Unexpected error", Some(requestId), retryable = false)
+    }
+
+  private def errorFromStatus(
+      resp: HttpResponse[String],
+      requestId: String,
+      defaultMessage: String,
+  ): ExtensionCallError =
+    ExtensionCallError(
```
jarekr-da, line 431: Low:
I see we do not use/log response body. Only status code.
During debug phase sometimes those messages are useful (BadRequest, 5xx( if server is configured to return diagnostic data).

Maybe logging at least some port of message makes sense?

community/participant/src/main/scala/com/digitalasset/canton/participant/extension/HttpExtensionServiceClient.scala
```diff
@@ -0,0 +1,527 @@
...
+      Some(requestId),
+      retryable = HttpExtensionServiceClient.isRetryableHttpStatus(resp.statusCode()),
+    )
+
+  private def shouldRetry(error: ExtensionCallError): Boolean =
+    error.retryable
+}
+
+object HttpExtensionServiceClient {
+  private[extension] val DefaultMaxResponseBodyBytes: Long = 20L * 1024L * 1024L
```
jarekr-da: Medium:
Maybe that should be configurable (per participant and/or per external server).
Also I do not see a test for this (both too large 200 and 5xx,4xx body)

community/participant/src/main/scala/com/digitalasset/canton/participant/extension/HttpExtensionServiceClient.scala
```diff
@@ -0,0 +1,527 @@
...
+        )
+
+      case None =>
+        try {
+          val requestBuilder = HttpRequest
+            .newBuilder()
+            .uri(endpoint)
+            .timeout(requestTimeout)
+            .header("Content-Type", "text/plain; charset=utf-8")
+            .header("Accept", "text/plain")
```
jarekr-da, line 293: Minor/Specification:

I do not remember if anywhere in spec we define maximum values for headers.
Might be a good idea to limit formally (1kb?) and validate here (to fail faster in case of errors).
Large headers can be blocked by proxies and it is better to be sure we detect problem earlier.

.proto_snapshot_image.bin.gz
jarekr-da, line 1: Is this accidentally commited?

community/participant/src/main/scala/com/digitalasset/canton/participant/extension/HttpExtensionServiceClient.scala
```diff
@@ -0,0 +1,527 @@
...
+  ): FutureUnlessShutdown[ExtensionValidationResult] =
+    startupLocalPreflight() match {
+      case Left(error) =>
+        FutureUnlessShutdown.pure(ExtensionValidationResult.Invalid(Seq(error)))
+
+      case Right(_) =>
+        validateVersionEndpoint()
+    }
+
+  private def validateVersionEndpoint()(implicit
```
jarekr-da, line 135: Documentation:

Maybe it makes sense to provide some formal spec for the server endpoint
Written doc is ok, and those are only 2 endpoints,
but formal spec like openapi can be used in case of misunderstandings.

jarekr-da (commented):
I put some comments for http specific parts.
Nothing critical - mostly things to consider.

community/participant/src/main/scala/com/digitalasset/canton/participant/extension/HttpExtensionServiceClient.scala
angelol: I added a capped debug log for non-200 response bodies, while keeping the body out of the caller-visible `ExtensionCallError`.

The logged diagnostic is skipped for empty bodies, capped at 1024 chars, and control-character sanitized. So it should help with debugging misconfigured services without changing the error surface or dumping unbounded response data into logs.

community/participant/src/main/scala/com/digitalasset/canton/participant/extension/HttpExtensionServiceClient.scala
angelol: I made this per-extension configurable via `ExtensionServiceConfig.maxResponseBodyBytes`, still defaulting to 20 MiB.

I also added focused coverage for oversized `200`, `400`, and `500` responses. In all three cases the client fails locally with `413`; the response body is not exposed in the returned error, and the oversized `500` is not retried because the local body-size limit takes precedence over HTTP retry classification.

community/participant/src/main/scala/com/digitalasset/canton/participant/extension/HttpExtensionServiceClient.scala
angelol: I added this in the HTTP client and in the service spec.

Canton now rejects outbound `X-Daml-External-Function-Id` and `X-Daml-External-Config-Hash` values longer than 1024 characters before sending the request, using the existing sanitized local `400` path. I kept `Authorization` out of this limit because bearer tokens can legitimately be larger and are validated separately.

The spec now states the same 1024 ASCII-character bound for non-auth protocol headers.

.proto_snapshot_image.bin.gz
angelol: This one is intentional. The branch changes `interactive_submission_data.proto`, and this file is the checked-in protobuf continuity snapshot generated for those descriptor changes.

It is easy to miss because the file is binary, but it is the standard `.proto_snapshot_image.bin.gz` update rather than an unrelated committed artifact.

community/participant/src/main/scala/com/digitalasset/canton/participant/extension/HttpExtensionServiceClient.scala
angelol: Agreed. We drafted a separate HTTP service specification for the external-call service contract and are using it for review now:

https://gist.github.com/angelol/4680ca242df586f82c9f6902cc7d9df2

For this PR I kept the Canton code change independent of the docs location, because the final documentation will likely live outside this repository. If reviewers prefer an OpenAPI artifact as a follow-up, we can derive one from this spec; the current draft is meant to pin down the endpoint behavior first.

.proto_snapshot_image.bin.gz
jarekr-da: ok, sorry

community/base/src/main/scala/com/digitalasset/canton/protocol/hash/NodeHashBuilderCommon.scala
```diff
@@ -101,15 +101,14 @@ private[hash] abstract class NodeHashBuilderCommon(
           exerciseResult,
           keyOpt,
           byKey,
-          externalCallResults,
+          _, // externalCallResults - intentionally excluded from the LF/prepared node hash.
+          // Prepared signatures authorize the LF action. External call result payloads are Canton
+          // protocol data, carried through ViewParticipantData.externalCallResults and protected
+          // by the Canton view signature.
```
andreaslochbihler-da: I don't understand this argument. What view signature are you referring to? 

AFAICT it would make sense to include external calls into the external party's signature because the external party does not sign the view hash. (The submitting participant signs the view hash.) Otherwise, the following attack is possible:

The external party uses an preparing participant node to translate the ledger API command into a prepared transaction. The party carefully inspects the prepared transaction to judge whether the transaction looks correct. If so, the party signs the transaction hash and forwards the prepared transaction with the hash to the submitting participant. The participant converts the prepared transaction into a transaction view tree and sends the confirmation request to the sequencer.

If the external call data is not signed by the external party, the preparing and submitting participant nodes can trick the party into signing a prepared transaction that contains the wrong external call results (because the party is not expected to re-execute the Daml interpretation), by showing the wrong external call data in the prepared transaction and swapping it out with the correct data before submission. If the party's decision to sign the transaction depends on checks that look at external call data, then the party would have signed something inadvertently.

This is the same reason why exercise results, for example, are part of the signature hash. Technically, the exercise results are completely determined by the other inputs if you have access to the Daml packages and a Daml engine. Since external parties typically do not, the exercise result must be signed. 

And by the same reasoning, external call data should be signed too.

community/base/src/main/scala/com/digitalasset/canton/protocol/hash/NodeHashBuilder.scala
```diff
@@ -99,12 +99,19 @@ private[hash] object NodeHashBuilder {
       : SortedMap[HashingSchemeVersion, SerializationVersion] =
     SortedMap(
       HashingSchemeVersion.V2 -> SerializationVersion.V1,
-      HashingSchemeVersion.V3 -> SerializationVersion.V2,
+      HashingSchemeVersion.V3 -> SerializationVersion.VDev,
```
andreaslochbihler-da: Serialization version V2 and HashingSchemeVersion V3 have been released with protocol version 35 in Canton 3.5. So they are frozen now and you'll have to add a new HashingSchemeVersion V4 that maps to SerializationVersion VDev.

community/base/src/main/scala/com/digitalasset/canton/protocol/hash/NodeHashBuilder.scala
```diff
@@ -99,12 +99,19 @@ private[hash] object NodeHashBuilder {
...
+      serializationVersion: SerializationVersion
+  ): Option[HashingSchemeVersion] =
+    HashingVersionToMaxSupportedLFSerializationVersionMapping.collectFirst {
+      case (hashingSchemeVersion, maximumSupportedLfVersion)
+          if Ordering[SerializationVersion].lteq(
+            serializationVersion,
+            maximumSupportedLfVersion,
+          ) =>
+        hashingSchemeVersion
+    }
```
andreaslochbihler-da, line 116: Is it intentional that you replace the one-time computation at class initialization with a lookup method that recomputes the value each time it is called?

community/ledger-api-proto/src/main/protobuf/com/daml/ledger/api/v2/interactive/transaction/v1/interactive_submission_data.proto
```diff
@@ -92,6 +122,14 @@ message Exercise {
   //
   // Required
   bool by_key = 16;
+  // Can only be populated from lf_version dev.
+  //
+  // Results of external calls made during this exercise.
+  // These results are carried for external-call replay and validation, but are
+  // not encoded into the external-signing prepared transaction hash.
```
andreaslochbihler-da: This probably has to be adapted given that the external call data should be covered by the hash.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
```diff
@@ -120,6 +131,162 @@ object DAMLe {
...
+  final case class ExternalCallRecordedResultDisagreement(
+      key: ExternalCallKey,
+      outputs: Set[LfBytes],
+  ) extends ReinterpretationError {
+    override protected def pretty: Pretty[ExternalCallRecordedResultDisagreement] = prettyOfClass(
+      param("extension id", _.key.extensionId.doubleQuoted),
+      param("function id", _.key.functionId.doubleQuoted),
+      param(
+        "config bytes",
+        disagreement => hexPayloadSize(disagreement.key.config).doubleQuoted,
```
andreaslochbihler-da, line 141: Minor: The pretty-printing here measures sizes in nibbles whereas the pretty-printing for `ExternalCallResultMismatch` measures them in bytes. Why the inconsistency?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
```diff
@@ -268,6 +268,64 @@ class ModelConformanceChecker(
...
+              externalCallResults.map(_.result)
+            )
+        }
+
+      logger.debug(
+        s"reInterpret: Aggregated ${storedExternalCallResults.size} external call results"
+      )
+
+      externalCallResultsByView
+        .parTraverse { case (currentView, externalCallResults) =>
```
andreaslochbihler-da: This `parTraverse` can lead to unbounded parallelism, which can cause issues (overloaded execution context, DB overload, etc.). Can you use `MonadUtil.parTraverseWithLimit(parallelism)` instead?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
```diff
@@ -268,6 +268,64 @@ class ModelConformanceChecker(
...
+      traceContext: TraceContext
+  ): FutureUnlessShutdown[ExternalCallReplayData] = {
+    val externalCallResultsByView =
+      if (!viewParticipantData.supportsExternalCallResults)
+        Seq.empty[(TransactionView, Seq[ViewParticipantData.ViewExternalCallResult])]
+      else
+        view.flatten.flatMap { currentView =>
+          currentView.viewParticipantData.unwrap match {
+            case Right(vpd) if vpd.externalCallResults.nonEmpty =>
+              Some(currentView -> vpd.externalCallResults.toSeq)
```
andreaslochbihler-da: Minor: Would it make sense to store the `externalCallResults` as a `Seq` in `ViewParticipantData` instead of as an `ImmArray`? I don't really see where we'd need the `ImmArray` in the first place.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
```diff
@@ -268,6 +268,64 @@ class ModelConformanceChecker(
...
+                      if result.checkingParties.intersect(hostedConfirmingParties).nonEmpty =>
+                    ExternalCallKey.fromResult(result.result)
+                }.toSeq
+          }
+        }
+        .map(validationKeyGroups =>
+          ExternalCallReplayData(
+            storedExternalCallResults = storedExternalCallResults,
+            validationKeyCounts =
+              validationKeyGroups.flatten.groupMapReduce(identity)(_ => 1)(_ + _),
```
andreaslochbihler-da: I do not really understand why we need these `validationKeyCounts` here. Naively, I would have expected that if a particular `ExternalCallKey` should be validated anywhere in the view, then it suffices to validate it exactly once. Here's my thought:

1. We assume that the external service is deterministic. So given the same inputs as defined by the `ExternalCallKey`, the external service always has to return the same output. In particular, the external service should in general not count how often it has been invoked because it's notoriously hard to agree on the number of invocations if there are retry loops anywhere (and just plain crash recovery of validation acts like such a loop).
2. If the recorded outputs for a given external call disagree, we anyway get a disagreement failure from the model conformance checker. So there is little point in trying to call the external service multiple times to see whether it's deterministic. This should be done as part of QA of the external service, not as part of the transaction protocol.
3. If you wanted to ensure that the validation is performed for exactly external calls in the subviews for which the participant hosts a confirming party, then this won't work anyway, because the view may have other subviews in between that perform the same external call key where the participant does not host a confirming party.

Can you explain and document this design choice here?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
```diff
@@ -268,6 +268,64 @@ class ModelConformanceChecker(
...
+            currentView.viewCommonData.tryUnwrap.viewConfirmationParameters.confirmers
+          topologySnapshot.canConfirm(participantId, confirmingParties).map {
+            hostedConfirmingParties =>
+              if (hostedConfirmingParties.isEmpty) Seq.empty
+              else
+                externalCallResults.iterator.collect {
+                  case result
+                      if result.checkingParties.intersect(hostedConfirmingParties).nonEmpty =>
+                    ExternalCallKey.fromResult(result.result)
+                }.toSeq
```
andreaslochbihler-da: Note that this logic overapproximates the validation rules that we had agreed on in the design phase for this feature. For reference, the validation rule that I remember is the following:

> An external call attached to an `Exercise` node must be validated by the signatories of the input contract (=`signatories` of the node) and the controllers of the choice (=`actingParties` of the nodes).

The confirming parties of a view are the union of all the confirming parties of the nodes in its core. So `confirmingParties` may include more parties than the `signatories` and `actingParties` of the `Exercise` node in the core of the `currentView`. For example, consider the following transaction:

* N1: Exercise node with signatory A and acting party B
  * N2: Exercise node with signatory A and acting party A, with a recorded external call result

Suppose that the informees of N2 are a subset of the informees of N1 (e.g., if there are neither contract nor choice observers). Then both nodes N1 and N2 belong to the core of the same view `V1` with confirming parties A and B, because B is a confirming party on `N1`.

By the above validation rule, only `A`'s participant node should have to validate the external call result because `A` is the only signatory or acting party of N2. Yet, by the logic implemented here, both `A`'s and `B`'s participant node would validate the external call.

This over-approximation may be problematic if `B`'s participant node does not even have the external call service configured. Then the transaction would not go through.

Since view decomposition depends on the party hosting relationship, I'd strongly discourage to rely on such over-approximations. Otherwise, a plain party replication process could break existing workflows because the view decomposition changes due to the replication and therefore the over-approximation changes.

andreaslochbihler-da (commented):
Here's a first round of comments. I didn't manage to look at everything yet, so I may still misunderstand something. I nevertheless wanted to share this feedback to shorten the iteration cycle.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
andreaslochbihler-da: Forget about this comment. This code does not over-approximate anything thanks to the intersection with `checkingParties`. What confused me was that you look at the confirming parties in the first place. The way that `checkingParties` is constructed ensures that they all are subsets of the confirming parties. IMO it would be clearer to completely ignore the confirming parties and instead compute the union of all checking parties in the external call results of the view participant data to look up with `topologySnapshot.canConfirm`. What's the reason you went with the confirming parties here?

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
```diff
@@ -323,12 +523,171 @@ class DAMLe(
...
+
+          case Right(outputO) =>
+            val (shouldValidate, stateAfterValidationDecision) =
+              stateWithReplayData.consumeValidation(externalCallKey)
+            outputO match {
+              case None =>
+                failExternalCall(ExternalCallReplayMissing(externalCallKey, currentCallIndex))
+
+              case Some(expectedOutput) if shouldValidate =>
+                validateExternalCallOutput(
```
andreaslochbihler-da: The error handling here can lead to ledger forks. AFAICT you do *not* want to interleave reinterpretation with validation of the external calls. Here's why:

* If reinterpretation fails with a `Left`, the model conformance checker treats the view itself as "obviously broken" and retries to reinterpret the subvies individually. See https://github.com/digital-asset/canton/blob/5ec182991db5d26c7c78920101467f3101ff6c11/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala#L166-L178
* For this reason, the checks during reinterpretation must be the same for all the participants that execute them, independently of what parties they host. Otherwise, some participants will conclude "OK" and others will conclude "bad".
* The mediator does not await the confirmation responses of all participants hosting any confirming party for liveness reasons. A verdict is taken as soon as a quorum of these participants has reached a conclusion. So the idea that "the participant has to confirm, therefore it can do more checks" falls short for multi-hosted parties.

Here is a concrete example in the context of external call:
* A transaction with a single node `N1` that consumingly exercises on an input contract `cid` with signatory `A` and contract observer `B` a choice `Ch` with controller `C` and choice observer `D`. So the confirming parties of this `Exercise` node (and thus view and transaction) are `A` and `C`.
* The body of `Ch` performs an external call `call`. The `checkingParties` parties are `A` and `C`.
* Party `C` is hosted on two participants `P1` and `P2` with confirmation threshold 1 and that party `B` is hosted on participants `P2` and `P3`. `A` is hosted on `P4` and `D` is hosted on `P5`. So participants `P1` and `P4` form a quorum for a approving the transaction, and so are `P2` and `P4`.

Suppose now that `P2` has not configured the extension required for the external call `call`. Since `P2` hosts the confirming party `C` with confirmation rights, it will attempt to validate `call`, but fail to do so. So `P2` sends a rejection to the mediator.

Yet, `P2` was a bit slow and `P1` and `P4` were faster in validating the transaction. Since they have correctly configured the extension, they have sent approvals. So overall the mediator has received a quorum of approvals first and therefore sent an approval to all participants. Here's what's going to happen:

1. `P1` and `P4` have validated the external call and found everything being fine. So they commit the transaction and expose it on their ledger APIs.
2. `P3` and `P5` do not host confirming parties. They simply used the recorded result for `call` in the reinterpretation and found everything is fine too. They also commit the transaction and expose it on their ledger APIs.
3. `P2`'s model conformance check has failed. So it decides to *not* commit the affected views, i.e., it rolls back the exercise and does not expose it on the ledger API. In particular, the input contract `cid` remains active from the perspective of `P2` whereas `P1`, `P3`, and `P4` think it's archived.

So we've ended up at a ledger fork.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallConsistencyChecker.scala
```diff
@@ -0,0 +1,160 @@
...
+    val empty: Result = Result(Map.empty)
+  }
+
+  /** Returns per-party inconsistencies for hosted confirming parties that can see disagreeing
+    * outputs for the same external call.
+    */
+  def check(
+      viewValidationResults: Map[ViewPosition, ViewValidationResult],
+      hostedConfirmingParties: Set[LfPartyId],
+  ): Result =
```
andreaslochbihler-da: Whenever there are disagreeing outputs for the same external call, this indicates that either the externally called service is broken or that the submitting participants is faulty. In either case, it would be good to raise a security alarm, independent of whether the participant hosts confirming parties.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallConsistencyChecker.scala
```diff
@@ -0,0 +1,160 @@
...
+              Iterator.single(viewPosition -> viewParticipantData.externalCallResults)
+            else Iterator.empty
+          }
+          .toSeq
+          .sortBy(_._1)(ViewPosition.orderViewPosition.toOrdering)
+
+      if (externalCallViewResults.isEmpty) Result.empty
+      else {
+        val outputsByPartyAndKey =
+          mutable.LinkedHashMap.empty[
```
andreaslochbihler-da: What's the point of using `LinkedHashMap`s here instead of a plain hashmap? Why do we care about the order in which the parties and inconsistencies are accumulated?

community/participant/src/main/scala/com/digitalasset/canton/participant/util/ExternalCallPayloadDescription.scala
```diff
@@ -0,0 +1,24 @@
...
+
+  def byteCount(bytes: Bytes): Int = bytes.toByteString.size()
+
+  def byteSize(bytes: Bytes): String =
+    s"${byteCount(bytes)} bytes"
+
+  def hexPayloadSize(hex: String): String =
+    Bytes
+      .fromString(hex)
+      .fold(_ => s"${hex.length} hex chars", byteSize)
```
andreaslochbihler-da: This looks odd. If `hex` is an invalid hex string of say 23 characters, we'll output `23 hex chars` (even though they are not hex chars), but if it is a valid hex string of say 24 nibles, then we output `12 bytes`.

community/participant/src/main/scala/com/digitalasset/canton/participant/ParticipantNode.scala
```diff
@@ -703,6 +708,38 @@ class ParticipantNodeBootstrap(
             )
             .mapK(FutureUnlessShutdown.outcomeK)
 
+        // Create extension service manager early so it can be shared with both CantonSyncService and LedgerApiServer
+        extensionServiceManagerO: Option[ExtensionServiceManager] =
+          if (parameters.engine.extensions.nonEmpty) {
```
andreaslochbihler-da: Minor: Consider to use `Option.when`.

community/participant/src/main/scala/com/digitalasset/canton/participant/ParticipantNode.scala
```diff
@@ -703,6 +708,38 @@ class ParticipantNodeBootstrap(
...
+            val manager = new ExtensionServiceManager(
+              extensionConfigs = parameters.engine.extensions,
```
andreaslochbihler-da, line 716: There does not seem to be any persistence as to what extensions have been configured. So if the configuration changes between two restarts of the participant, a participant may actually send conflicting responses to the mediator for the same confirmation request when crash recovery reprocesses such a not-yet-decided confirmation request whose external calls are affected by the configuration change.

This makes the participant look faulty to the synchronizer. It might make sense to rely on a three-valued approval logic in general: If the participant hosts a checking party for an external call with confirmation rights, but the external call cannot be validated because the extension service is not configured, then the participant should respond with `Abstain` instead of `Reject` (clearly only if everything else is fine with the confirmation request). In contrast, if the external serivce is configured, but validation fails because the external service produces a different output, then a rejection remains to be the appropriate confirmation response.

This idea relies on the assumption that it is more likely that someone misconfigures the external services than that the external service itself is buggy and produces non-deterministic outputs.

andreaslochbihler-da (commented):
I've left another round of high-level feedback on the approach. AFAICS the wiring needs to be changed quite a bit to avoid ledger forks.

matthiasS-da:
As already pointed out on Slack, I'll taking over the review from @andreaslochbihler-da. Happy to work with you!

community/base/src/main/scala/com/digitalasset/canton/protocol/hash/NodeHashBuilder.scala
matthiasS-da: I guess the intention is to provide a HashingSchemeVersion even if the serializationVersion does not occur literally in the map. The method is only called to produce error messages. So I don't see a problem with that.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallConsistencyChecker.scala
matthiasS-da: Look at SyncServiceAlarm for prior art, but please create your own, as the resolution "contact support" is not appropriate.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -29,6 +47,390 @@ class TransactionConfirmationResponsesFactory(
...
+    checked(
+      ConfirmationResponse
+        .tryCreate(
+          Some(viewPosition),
+          localVerdict,
+          parties,
+        )
+    )
+
+  private def ordinaryLocalVerdictAndParties(
```
matthiasS-da, line 159: My understanding is that the changes starting here are motivated by sending rejections on inconsistencies of external calls.
I'm a bit surprised by the amount of changes.

What are you trying to achieve here?
What has motivated the change of design?

community/base/src/main/scala/com/digitalasset/canton/crypto/InteractiveSubmission.scala
```diff
@@ -129,9 +142,22 @@ object InteractiveSubmission {
...
+        )
+      if (protocolVersion < minimumProtocolVersion)
+        Left(
+          UnsupportedLfSerializationVersion(
+            transaction.version,
+            protocolVersion,
+            minimumProtocolVersion,
+          )
+        )
+      else
```
matthiasS-da: Note that this bit is called by ModelConformanceChecker and therefore this is a change of released code running as part of Phase 3. As such, the change may be breaking and can cause ledger forks. Have you already checked that it does not change the behavior of ModelConformanceChecker, and if so why not?

community/base/src/main/scala/com/digitalasset/canton/protocol/LocalRejectError.scala
```diff
@@ -116,6 +116,23 @@ object LocalRejectError extends LocalRejectionGroup {
             _resourcesType = Some(ErrorResource.ContractId),
           )
     }
+
+    @Explanation(
+      """This rejection is made by a participant if external call results visible to a confirming
+        |party disagree on the output for the same external call.
+        |"""
+    )
+    @Resolution("This indicates either malicious or faulty behaviour.")
```
matthiasS-da: My understanding is that this kind of error can happen in production (either due to a malicious submitting participant or due to operational problems with external calls). Therefore, we need a more descriptive resolution here.

community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceExternalCallHandler.scala
```diff
@@ -0,0 +1,64 @@
...
+        mode,
+      )
+      .map(_.left.map { extensionError =>
+        ResultNeedExternalCall.Error(sanitizedEngineError(extensionError))
+      })
+
+  private def sanitizedEngineError(extensionError: ExtensionCallError): String = {
+    val requestId = extensionError.requestId.fold("")(requestId => s", requestId=$requestId")
+    s"External call failed with status ${extensionError.statusCode}$requestId"
+  }
```
matthiasS-da, line 45: I realize that you're sanitizing quite aggressively, i.e., discard the error message, keep only the status code. What is the underlying strategy? How do you balance security with a good operational experience?

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
```diff
@@ -323,12 +523,171 @@ class DAMLe(
...
+    def validateExternalCallOutput(
+        extensionId: String,
+        functionId: String,
+        configHash: String,
+        input: String,
+        expectedOutput: LfBytes,
+        externalCallKey: ExternalCallKey,
+        currentCallIndex: Int,
+        resume: Either[ResultNeedExternalCall.Error, String] => Result[A],
+        externalCallState: ExternalCallLoopState,
```
matthiasS-da: This parameter list seems to be redundant:
- externalCallKey -> extensionId, functionId, configHash, input
- externalCallState -> currentCallIndex

I kindly ask you to avoid such redundancy.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
```diff
@@ -323,12 +523,171 @@ class DAMLe(
...
+                  configHash,
+                  input,
+                  expectedOutput,
+                  externalCallKey,
+                  currentCallIndex,
+                  resume,
+                  stateAfterValidationDecision,
+                )
+
+              case Some(storedOutput) =>
```
matthiasS-da: This is somewhat related to what Andreas said to validationKeyCounts.
Naively, I would just have expected that every output needs to be validated. Here you have some logic that skips validation once validationKeyCounts have been exceeded. That creates the (at least theoretical) risk of skipping a necessary validation of external call outputs. 
Could you please explain why you decided to sometimes skip the validation and why this does not introduce a vulnerability?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
matthiasS-da: Related to this, consider to introduce a size limit on external call results to prevent DoS attacks.

matthiasS-da:
Before the weekend, let me summarize the current state.

We have identified several critical issues:

1. Security: External call results are [excluded from the transaction hash](https://github.com/digital-asset/canton/pull/541#discussion_r3465733654), meaning they aren't covered by the external party's signature. An attacker can therefore tamper with the call results.
2. Breaking Changes: The serialization version / hashing scheme version mapping in NodeHashBuilder.scala has been [changed on a released version](https://github.com/digital-asset/canton/pull/541#discussion_r3465761954). A new hashing scheme needs to be introduced instead.
3. Ledger Fork: External call outputs are [validated during the model conformance check](https://github.com/digital-asset/canton/pull/541#discussion_r3472842376). If validation results differ among participants, it will trigger an immediate ledger fork.

Furthermore, we have identified the following areas of concerns:

1. Potential DoS vulnerability due to [use of parTraverse on untrusted data](https://github.com/digital-asset/canton/pull/541#discussion_r3466645366).
2. An honest participant gets [flagged as malicious by mediators](https://github.com/digital-asset/canton/pull/541#discussion_r3473012370).
3. Risk of [skipping the validation of external call outputs](https://github.com/digital-asset/canton/pull/541#discussion_r3480186266) due to code complexity. To be discussed.
4. Another [potentially breaking change](https://github.com/digital-asset/canton/pull/541#discussion_r3475635774).
5. Concerns around code complexity, e.g., [here](https://github.com/digital-asset/canton/pull/541#discussion_r3466766237), [here](https://github.com/digital-asset/canton/pull/541#discussion_r3472293045), [here](https://github.com/digital-asset/canton/pull/541#discussion_r3475596394), [here](https://github.com/digital-asset/canton/pull/541#discussion_r3480090533). To be discussed.
6. Concerns around error handling, e.g., [here](https://github.com/digital-asset/canton/pull/541#discussion_r3472907680), [here](https://github.com/digital-asset/canton/pull/541#discussion_r3475645362), [here](https://github.com/digital-asset/canton/pull/541#discussion_r3475661309). To be discussed.

As these issues would lead to security vulnerabilites, ledger forks, or severe code complexity, we cannot merge the code in its current state.

Reviewing a PR of this footprint also poses a massive resource challenge for our team. We do not accept single PRs of this size even from internal developers, and we must kindly ask you to follow our standard practice of providing smaller, incremental PRs of no more than 1000 lines (as a soft limit). To make this manageable, we suggest splitting the work into the following steps:

1. Get #537 approved and merged. Then rebase this PR.
2. Define a new hashing scheme for externally signed transactions.
3. Overhaul the submission path: connect StoreBackedCommandInterpreter with the external call handler.
4. Overhaul the validation path: connect DAMLe with the external call handler.
5. Check consistency of the recorded external call outputs and extend TransactionConfirmationResponseFactory.

We genuinely appreciate the massive amount of effort you've put into this feature, but we need to prioritize the stability of the protocol. Please take your time to look through these points, and let us know how you would like to proceed with restructuring the PRs once you've had a chance to review them.

community/participant/src/main/scala/com/digitalasset/canton/participant/ParticipantNode.scala
angelol: Done, switched to `Option.when`.

community/base/src/main/scala/com/digitalasset/canton/crypto/InteractiveSubmission.scala
angelol: Good catch, this needed a code change, not just an answer. That guard was added by this PR; it isn't on `main`. I removed it, so `InteractiveSubmission.scala` is now identical to `main` and behaves exactly as released on the Phase-3 path. It was redundant anyway: the same boundary is already enforced by protocol-version → supported hash schemes, composed with hash scheme → max LF version (asserted per node in `NodeHashBuilder`). The only hashing change left versus `main` is the new, dev-only `V4`; V2 and V3 are untouched and the existing v2/v3 golden hash tests pass.

community/base/src/main/scala/com/digitalasset/canton/protocol/LocalRejectError.scala
angelol: Agreed. I've expanded the resolution to name the likely production causes (a faulty or malicious submitter, inconsistent external-call service configuration, or a service that returns unstable output for the same call) and to point the operator at the submitting participant and the service deployments. The specific call identity and conflicting outputs are already in the rejection's cause.

community/base/src/main/scala/com/digitalasset/canton/protocol/hash/NodeHashBuilder.scala
angelol: Done. I reverted V3 to its frozen 3.5 mapping (`V3 → V2`) and added the new `V4 → VDev` as you suggested. V4 is dev-protocol-only (v4 node/metadata hashers, proto v32, and the ledger-API enum wired through), and it's where development-version `external_call_results` are now folded into the prepared-transaction hash.

community/base/src/main/scala/com/digitalasset/canton/protocol/hash/NodeHashBuilder.scala
angelol: Yes, intentional. Unlike the old precomputed map, which only resolved serialization versions that appear verbatim as values, this returns a supporting scheme for any serialization version via the `<=` comparison. And as you note, it's only used to build one error message in `TransactionHash`, so the per-call cost doesn't matter.

community/base/src/main/scala/com/digitalasset/canton/protocol/hash/NodeHashBuilderCommon.scala
angelol: You're right, and thanks for the worked example. I've moved external-call results into the signed hash: the new V4 hashing scheme encodes each result (extension id, function id, config, input, output) into the prepared-transaction hash. It can't be bypassed, results are only allowed on dev-version nodes (the decoder rejects them otherwise), and a dev-version node can only be hashed by V4 since V2/V3 reject that serialization version. So a preparing/submitting participant can no longer swap results after signing: validation recomputes the V4 hash over the submitted results and the external signature fails if they differ.

community/ledger-api-proto/src/main/protobuf/com/daml/ledger/api/v2/interactive/transaction/v1/interactive_submission_data.proto
angelol: Done. The comment now says the results are encoded into the prepared-transaction hash starting with `HASHING_SCHEME_VERSION_V4`, replacing the previous "not encoded into the hash" note.

community/participant/src/main/scala/com/digitalasset/canton/participant/ParticipantNode.scala
angelol: This is deliberate: I didn't persist the validation outcome. External-call results are required to be deterministic, and validation re-executes the call and checks the live service against the output recorded in the externally-signed transaction. On recovery the participant re-runs that check against the current config, so a deterministic service (including one that simply moved to a new endpoint) returns the recorded output again and the response is unchanged. If the service no longer reproduces the recorded output, rejecting is the intended outcome, and persisting the old result would mask exactly that. And if the extension has been removed so the call can't be validated at all, the participant abstains (your three-valued suggestion) and defers to the mediator rather than rejecting. So config drift is handled by re-validating against the current config, not by freezing per-request state, and the participant's commit follows the mediator verdict regardless.

community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceExternalCallHandler.scala
angelol: The strategy is to keep the cross-trust-boundary surface minimal and the local surface rich. The external service is across a trust boundary, so its error body can carry sensitive internals or attacker-influenced content, and I don't let it flow into a confirmation response: the abstain reason that reaches the mediator and other nodes carries only the numeric HTTP status plus an opaque request id, never the service message. The body isn't lost, just kept local: `HttpExtensionServiceClient` logs it at debug level, capped at 1024 chars and control-character sanitized. The request id appears in both the abstain reason and that log, so an operator can correlate the cross-node abstain with the local body and the service's own logs. So status plus request id are enough to triage (4xx vs 5xx) and correlate without leaking across the trust boundary, and the full body is one debug level away on the node itself.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallConsistencyChecker.scala
angelol: Done. I added a dedicated `ExternalCallResultDisagreementAlarm` (an `AlarmErrorCode` with the same shape as `SyncServiceAlarm`) and gave it an actionable resolution pointing at the submitting participant and the external-call service deployments, rather than "contact support". It fires off the visible recorded disagreements, which are computed independently of hosted parties, so it alarms even when this participant hosts no confirming party (including on the malformed-payload path).

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallConsistencyChecker.scala
angelol: We do care about output order, for deterministic confirmation responses and alarms (the responses get re-sent on crash recovery, so they need to be reproducible), but not the accumulation order. The LinkedHashMaps were an implicit way to get that ordering via insertion order. I've removed them: the checker now uses plain maps and applies explicit `Ordering`s (`orderInconsistency`, `orderOutputSets`) where the output is actually observable, so determinism no longer rides on map iteration order.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
angelol: Done. `externalCallResults` is now a `Seq` in `ViewParticipantData` (the `ImmArray` and the `.toSeq` conversions it forced are gone), and the tree factories were updated to match.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
angelol: I checked, and external-call results are already bounded here: they ride inside `ViewParticipantData` in a confirmation request, which is capped by `maxRequestSize`, so the work is linear in a bounded input (and disagreement diagnostics report sizes only, truncated to 1024 chars). While checking I did find one avoidable allocation, materializing outputs as hex just to sort them, which I've fixed to order by raw bytes instead. So I don't think a dedicated limit is needed, but happy to add a per-result cap as defense-in-depth if you'd prefer.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
angelol: You're right that the `checkingParties` intersection keeps it from over-approximating and that `canConfirm(checkingParties)` would select the same set. That code has since been removed: to fix the ledger fork you flagged on DAMLe.scala:663, validation moved out of the reinterpretation loop into the confirmation-response phase. The party-selection there now keys off the view's already-approving parties, which reuses the `canConfirm` the verdict already did and only re-executes a call for a party we'd otherwise approve rather than one we're already rejecting.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
angelol: You're right on all three, and I removed `validationKeyCounts` entirely. Validation no longer runs in the reinterpretation loop (it moved to the confirmation-response phase, see the DAMLe.scala:663 thread), and there it runs once per distinct `ExternalCallKey`, not per occurrence and not counted. The service is required to be deterministic, so one call per key answers the only question that matters; dropping the count also removes the hazard you flagged, since with a count the number of service calls would depend on how often crash recovery or the subview retry re-runs, which differs across participants. Recorded-output disagreements are caught by the consistency check (and already fail replay) rather than by re-calling the service, and keying validation by `ExternalCallKey` rather than by subview handles the interleaved-subview case.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: You're right it was heavier than it needed to be. I've lifted the external-call routing and validation out into a dedicated `ExternalCallResponseRouter` (matching the package's existing `*Checker`/`*Validator` split) and trimmed the incidental churn, so the factory is back to a thin response-assembler (~890 → ~520 lines), behaviour unchanged.

What's genuinely structural: results are checked by specific parties, so a view splits into per-party reject/abstain/approve instead of one verdict; the live re-execution runs in the confirmation-response phase rather than reinterpretation, to avoid the ledger fork Andreas raised on DAMLe.scala:663; and disagreements are detected both per-root-view (during replay) and by a global scan across views (which also alarms regardless of hosting), because neither scope subsumes the other.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
angelol: The divergent one is gone: `ExternalCallResultMismatch`, which sized its raw `LfBytes` with `byteSize`, was removed when validation moved out of reinterpretation. The remaining errors now report sizes uniformly in bytes, config/input are engine-emitted canonical hex and `hexPayloadSize` decodes them to a byte count, matching the outputs' `byteSize`. The only non-byte path is the defensive fallback for non-canonical input, which now reports "input characters" (your other comment) and shouldn't occur for engine hex.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
angelol: Removed. The method with the redundant list (`validateExternalCallOutput`) is gone with the validation relocation. I also went further on the second point you raised: the per-call index and the loop-state object it lived in are gone entirely — they only fed a diagnostic and were never used for matching (replay matches purely on `ExternalCallKey`). The replay path now takes just the `ExternalCallKey` plus the memoized replay data, so there's no derived-and-also-passed index left.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
angelol: You're right, this was a real fork risk. I've taken external-call validation out of reinterpretation entirely: `DAMLe` now only replays recorded outputs (the handler parameter and `validateExternalCallOutput` are gone), so reinterpretation is identical on every participant. Re-execution now happens in the confirmation-response phase, where a hosted checking party whose call can't be validated abstains instead of rejecting (per your `ParticipantNode.scala:715` note). So the participant in your example no longer rolls back: it abstains and defers to the mediator. Covered by the abstain / mismatch / malformed cases in `TransactionConfirmationResponsesFactoryExternalCallTest`.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
angelol: There's no count-based skip anymore, `validationKeyCounts` is gone (the removal Andreas asked about on `ModelConformanceChecker`). Each distinct `ExternalCallKey` with a single recorded output is validated exactly once. The only keys not re-executed are those whose recorded outputs already disagree, and a disagreement already fails replay and is rejected by the consistency check, so re-executing them would add nothing. So nothing that needs validating is skipped.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/ExternalCallPayloadDescription.scala
angelol: Fixed the label: the fallback now says "input characters" instead of "hex chars", so non-hex input isn't called "hex chars". Valid (canonical) hex still reports the decoded byte count. The remaining char-vs-byte difference is intentional, an unparseable string has no byte length to report, so we give its raw length honestly labeled, and config/input are engine-emitted canonical hex so the fallback shouldn't normally fire.

angelol:
Thanks Matthias — really helpful summary, and the size/stability concerns are fair. I've pushed fixes for all three critical issues and replied inline on each thread; at the PR level:

1. **Hash coverage / tampering** — external-call results are now folded into the prepared-transaction hash via a new dev-only `HashingSchemeVersion.V4`, so they're covered by the external party's signature and can't be swapped after signing.
2. **Breaking change** — I reverted the released V2/V3 mapping to its frozen 3.5 state and added `V4 → VDev` instead, as you suggested; V2/V3 are untouched and their golden-hash tests still pass.
3. **Ledger fork** — external-call validation is out of the reinterpretation / model-conformance path entirely, so reinterpretation only replays recorded outputs and is identical on every participant. Re-execution moved to the confirmation-response phase, where a hosted checking party that can't validate **abstains** instead of rejecting — so the fork scenario you described can't occur.

The secondary concerns are addressed or answered in their threads too — bounded parallelism plus the existing `maxRequestSize` bound (DoS), abstain-not-reject (mediator flagging), validation once per `ExternalCallKey` with `validationKeyCounts` removed, and the factory split into a dedicated `ExternalCallResponseRouter` for the complexity concern. One open offer: a per-result size cap as extra defense-in-depth if you'd like it. Happy to dig further into the "to be discussed" items.

On the footprint — one bit of context first: GitHub reports ~7.6k changed lines because this PR targets `main` while it's stacked on #537 (the 05 redesign), which isn't merged yet, so the count includes all of #537. The delta unique to *this* PR on top of #537 is ~5.7k lines across 85 files, and of that only ~2.1k is production Scala — the rest is ~3.4k of tests plus generated proto code, the descriptor snapshot, and docs.

If that ~2.1k production surface is still over your bar, I'm glad to split — but I'd suggest splitting the PR *as it stands now*, fixes included, into a few smaller self-contained PRs, rather than rebuilding it through steps 2–5. The feature was developed as a single unit, so reproducing it as a forward sequence would mean intermediate PRs that aren't independently mergeable until a later one completes them; slicing the finished, fixed code instead keeps every PR in good shape. Your layer boundaries are still the natural seams to cut along (hashing scheme / submission path / validation path / consistency + factory), so it's mainly reshuffling on our side, which we're happy to do. Either way #537 lands and we rebase first.

A couple of things to confirm: would that approach — slicing the current fixed PR along those seams — work for you? And does the ~1000-line soft limit count test code or production only? (At ~2.1k production vs ~3.4k tests, that materially changes how many slices we're looking at.)

Thanks again for the thorough review.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
matthiasS-da: We are aware of `maxRequestSize` - however the original concern would have applied nevertheless. `maxRequestSize` needs to cover the payload of the entire requests and may need to be configured generously for unrelated reasons to up to 2GB. You can fit many external calls into that limit; depending what the code does with the external calls, resource exhaustion can still occur.

With that, I dedicated size limit will likely be needed. Let me check how to best get this in place. I'll get back to you.

community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceExternalCallHandler.scala
matthiasS-da: Thanks for the explanation. I certainly need to review this again.
Two points that caught my immediate attention:
- You can't assume that DEBUG logging is enabled in production. The level probably needs to be raised to INFO.
- Capping at 1024 is nice for keeping logs concise, but can omit important information. So the size limit may have to be revisited.

matthiasS-da:
@angelol thanks for your comprehensive reply! Very much appreciated!

For the sake of clarity, the 1000 line soft limit refers to all code changes (production, test, documentation). It only excludes whitespace changes and lines changed due to moving code around.

The approach of how to split this into separate PRs is totally up to you. Take my input on this as a suggestion. Also, I understand that it makes sense to create one big branch/PR up-front to verify that the pieces fit together.

I mostly agree with your actions / replies on the detailed comment. I've left a note in the cases where I didn't.

Looking forward to receive the first splitted PR to review.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
matthiasS-da: For the time being I can offer that we'll introduce the limit for you. (I may have to take back the offer in a couple of weeks, but let's optimistically assume that your changes will land before that.)

community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceExternalCallHandler.scala
angelol: Done on both points:
- Raised the response-body diagnostic from DEBUG to **INFO**, so it's available in production where DEBUG is typically off. It's still on the failure path only (built in `errorFromStatus`), so the extra volume is bounded to non-2xx responses.
- Raised the logged-excerpt cap from 1024 to **8192** chars, enough for a full error message plus a meaningful stack-trace excerpt.

The security posture is unchanged: the body stays local to the node (never enters a cross-node confirmation response), is still control-character sanitized, and the excerpt is still bounded (the full body is already size-limited by `maxResponseBodyBytes`). Happy to adjust the level or cap further once you've had another look.

PR #549, opened 2026-06-30
feat(external-call): protocol and transaction-hash foundation

angelol:
## Summary

This is PR 1 of 8 in the external-call runtime-integration series tracked in #513, building on #537 (now merged to `main`).

This PR lays the protocol and transaction-hash foundation the rest of the series builds on: the Canton protocol fields that carry recorded external-call results, and the `HashingSchemeVersion` V4 hashing that covers them.

## Scope

This PR adds:

- `ViewParticipantData.externalCallResults` and the supporting protocol additions (`participant_transaction.proto`, `SubmitterMetadata`, `ExternalAuthorization`, `LocalRejectError`), including the `ImmArray → Seq` migration of the recorded-results collection across the base/common/participant transaction-tree construction sites
- `HashingSchemeVersion` V4 (dev-gated) and the v4 node/transaction hashers that fold recorded external-call result payloads into the transaction hash, with the payloads surfaced in the `HashTracer` output
- the `HASHING_SCHEME_VERSION_V4` Ledger API enum value and the matching `SubmitRequestValidator` case (required for `-Werror` match exhaustiveness once the enum exists)
- the daml-lf `transaction.proto` field for recorded results
- the regenerated protobuf snapshot and `openapi.yaml` (V4 enum), plus hash tests, generators and docs

## Out of scope

The prepared-transaction codec (PR 2), command-execution wiring (PR 3), engine replay (PR 4), the extension-service runtime (PR 5+), consistency checking, routing and validation (PR 6–8). No LF semantics change. The feature is `ProtocolVersion.dev`-gated, so this PR is dormant on released protocol versions.

## Hashing Boundary

Recorded external-call result payloads are Canton protocol data: after submission they are carried in `ViewParticipantData.externalCallResults` and protected by the Canton view signature. V4 hashing folds them into the Canton transaction hash on that basis. (Their exclusion from the LF/prepared node hash is covered in PR 2.)

## Stacking & review

PR 1 of an 8-PR series; it is the base, so its diff against `main` is already the clean per-PR change. The series is stacked, so PRs 2–8 should be reviewed/merged in order (each rebased onto `main` as the previous lands). Dependency notes are in each PR.

Refs #513.


community/base/src/main/scala/com/digitalasset/canton/protocol/hash/v4/NodeHashBuilder.scala
```diff
@@ -0,0 +1,51 @@
...
+      enforceNodeSeedForCreateNodes,
+    ) {
+
+  override protected val hashingSchemeVersion: HashingSchemeVersion = HashingSchemeVersion.V4
+
+  override private[hash] def newBuilder(hashTracer: HashTracer): NodeHashBuilder =
+    new NodeHashBuilder(purpose, hashTracer, enforceNodeSeedForCreateNodes)
+
+  private def addExternalCallResult(result: ExternalCallResult): this.type =
+    addContext("External Call Result")
```
matthiasS-da, line 38: Please use `val ExternalCallResult(...) = result` to retrieve the fields of result. That way, it is easier to see when a field is excluded from the hash.

community/base/src/main/scala/com/digitalasset/canton/protocol/hash/NodeHashBuilderCommon.scala
```diff
@@ -101,15 +101,12 @@ private[hash] abstract class NodeHashBuilderCommon(
           exerciseResult,
           keyOpt,
           byKey,
-          externalCallResults,
+          _, // externalCallResults - not part of the V2/V3 common exercise-node encoding.
```
matthiasS-da: Please continue failing on hashing scheme versions v1-v3, if external call results are present!

community/base/src/main/scala/com/digitalasset/canton/protocol/LocalRejectError.scala
```diff
@@ -116,6 +116,25 @@ object LocalRejectError extends LocalRejectionGroup {
             _resourcesType = Some(ErrorResource.ContractId),
           )
     }
+
+    @Explanation(
+      """This rejection is made by a participant if external call results visible to a confirming
+        |party disagree on the output for the same external call."""
+    )
+    @Resolution(
+      "Inspect the submitting participant and the external-call service deployments. The same external call was recorded with different outputs, which can indicate a faulty or malicious submitter, inconsistent external-call service configuration, or a service that does not return a stable output for the same external-call identity."
```
matthiasS-da: The text is much better!

MINOR: Please break up the line to max 120 characters. You can use
"""...
   |..."""
but you need to omit the usual .stripMargin.

community/daml-lf/transaction/src/main/protobuf/com/digitalasset/daml/lf/transaction.proto
```diff
@@ -127,7 +127,7 @@ message KeyWithMaintainers {
 message ExternalCallResult {
   string extension_id = 1; // Extension service identifier (e.g., "oracle-price-feed")
   string function_id = 2; // Function identifier within the extension
-  bytes config = 3; // Extension configuration (binary)
+  bytes config = 3; // Extension configuration hash bytes
```
matthiasS-da, line 130: I don't get this. TransactionCoder.Internal encodes all binary fields the same. Why is config "hash bytes", whereas output is "binary"?

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
```diff
@@ -43,34 +43,50 @@ The hashing algorithm is tied to the protocol version of the synchronizer used t
...
-==================  =========================
-v34                 V2
-v35                 V2, V3
-==================  =========================
+====================  =========================
+Protocol Version      Supported Hashing Schemes
+====================  =========================
+v34                   V2
+v35                   V2, V3
+development protocol  V2, V3, V4
```
matthiasS-da: Please replace "development protocol" by "dev".

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
```diff
@@ -43,34 +43,50 @@ The hashing algorithm is tied to the protocol version of the synchronizer used t
...
 
 Transaction Nodes
 -----------------
 
 Transaction nodes are additionally individually versioned with a Daml version (also called LF version).
 The encoding version is decoupled from the LF version and implementations should only focus on the hashing version.
 However, new LF versions may introduce new fields in nodes or new node types. For that reason, the protobuf representation of a node is
-versioned to accommodate those future changes. In practice, every new Daml language version results in a new hashing version.
+versioned to accommodate those future changes. In stable protocol versions, new Daml language versions may result in new hashing
+versions. Development-only Daml versions may be supported under the development protocol before a stable hashing version is assigned.
```
matthiasS-da: What was the reason to change this?
Why does it say "develoment-only" Daml versions (plural)? AFAIK, there is only one development Daml version.
What is the "development protocol"?

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
```diff
@@ -43,34 +43,50 @@ The hashing algorithm is tied to the protocol version of the synchronizer used t
...
+
+General approach
+----------------
+
+V4 follows the same general hashing approach as V3. The V3 section below defines the common encoding used by V4 unless a V4-specific difference is stated.
+
+Changes from V3
+---------------
+
+- Development-version Exercise nodes include ``external_call_results`` in the prepared transaction hash.
```
matthiasS-da: MINOR: Please write "Exercise nodes with version dev include ..."

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
```diff
@@ -91,6 +107,8 @@ Changes from V2
 
 - Addition of an ``max_record_time`` field in :ref:`metadata <metadata_encoding>` to make maximum record time explicit in the signed metadata.
 
+V3 is the released PV35 hashing scheme for LF ``SerializationVersion.V2``. It remains stable and does not include development-version ``external_call_results``.
```
matthiasS-da: I agree with the statement, but why do you please this comment in this section? How is this a change from V2?

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
```diff
@@ -104,7 +122,7 @@ Changes from V1
 
     V3 introduces support for contract keys. Usage of contract keys in externally signed transactions
     requires usage of V3. Contract keys will not work on V2.
-    Also note that V3 is only :ref:`supported <hashing_scheme_version>` on protocol version 35.
+    Also note that V3 is the stable hashing scheme introduced at protocol version 35, and V4 is :ref:`supported <hashing_scheme_version>` only on the development protocol.
```
matthiasS-da: Why do you make statements about V4 in the section about V3?

community/base/src/main/scala/com/digitalasset/canton/protocol/hash/v4/NodeHashBuilder.scala
angelol: Done in 4fa6a729b — `addExternalCallResult` now binds every field via `val ExternalCallResult(extensionId, functionId, config, input, output) = result`, so adding a field to `ExternalCallResult` stops the build and forces a conscious hashing decision. The hash value is unchanged (the v4 `NodeHashTest` is still green).

community/base/src/main/scala/com/digitalasset/canton/protocol/hash/NodeHashBuilderCommon.scala
angelol: Done in 6d72dd5ff — the shared V2/V3 exercise-node encoding now fails with `NodeHashingError.UnsupportedFeature` when `externalCallResults` is present, instead of silently dropping them.

Implemented via a virtual `supportsExternalCallResults` hook (default `false`; V4 overrides to `true`), mirroring `ViewParticipantData.supportsExternalCallResults`. I deliberately avoided reading `hashingSchemeVersion` here: in `NodeHashBuilderCommon` that name resolves to the constructor parameter (`V3`) even for a V4 instance, so a version comparison would wrongly reject V4. Added a test asserting pre-V4 schemes reject results-bearing exercise nodes.

community/base/src/main/scala/com/digitalasset/canton/protocol/LocalRejectError.scala
angelol: Done in 7e036a1e7 — broke it into a `|`-margin multi-line string (no explicit `.stripMargin`, matching the `@Explanation` above it); all lines are now ≤100 chars.

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
angelol: Done in 3b78fd215 — the table now reads `dev`.

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
angelol: Done in e9b715696 — reworded to "Exercise nodes with version dev include ...".

community/daml-lf/transaction/src/main/protobuf/com/digitalasset/daml/lf/transaction.proto
angelol: You're right that `TransactionCoder.Internal` encodes all three `bytes` fields identically — the comments are describing each field's semantic role, not its encoding.

`config` carries the **configuration hash** that Daml supplies for the call: a digest, not the raw configuration. The raw config is never transmitted — the hash is opaque, with no fixed algorithm or digest length. `input` and `output` are the raw function payload bytes.

This is defined in the external-call HTTP service spec (you may not have seen it — I shared it with the team as a gist): https://gist.github.com/angelol/4680ca242df586f82c9f6902cc7d9df2

See the *External-Call Model* — `(extension id, function id, config hash, input bytes) -> output bytes` — and the `X-Daml-External-Config-Hash` section.

So the distinction is intentional: `config` is a hash, whereas `input`/`output` are raw bytes. Happy to reword the comments if you'd find it clearer.

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
angelol: You're right — that change wasn't well-founded and the wording was imprecise. Reverted to the original sentence in e32641d6b. The mapping is simply that each Daml/LF version maps to a new hashing version (LF V1→V2, V2→V3, dev→V4).

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
angelol: Agreed — that isn't a change from V2, so I removed it in 8102b587f. V3's protocol-version support is in the table above, and the `external_call_results` detail is covered in the V4 section.

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
angelol: Agreed — V4 doesn't belong in the V3 section (it has its own section now), so I dropped it in 1e0c5772f. I also removed "only": per `HashingSchemeVersion`, V3 is supported from v35 onwards (including dev), not v35 alone.

community/base/src/main/scala/com/digitalasset/canton/protocol/hash/NodeHashBuilderCommon.scala
angelol: Update: following @thibault-da's suggestion on #550, this guard now keys on the LF serialization version (`version != SerializationVersion.VDev`) — mirroring `PreparedTransactionDecoder` — rather than the scheme hook. V2/V3 still reject external-call results as you asked (they only ever hash pre-dev nodes). Commit 993b94e47.

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
```diff
@@ -647,6 +665,8 @@ Exercise
         encode(exercise.consuming) ||
         encode(exercise.exercise_result) ||
         encode(exercise.choice_observers) ||
+        encode(exercise.by_key) ||
+        encode(exercise.key) ||
```
matthiasS-da: Thanks!

matthiasS-da: @paulbrauner-da This has been forgotten when keys have been added to the hash function. I think a definition of `encode(exercise.key)` is missing. Could you please take care?

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
```diff
@@ -759,13 +814,27 @@ Final Hash
...
-        0x03 || # Hashing Scheme Version
-        hash(transaction) ||
-        hash(metadata)
+      .. code-block::
+
+          fn encode(prepared_transaction):
+              0x00000030 || # Hash purpose
+              0x04 || # Hashing Scheme Version
+              hash(transaction) ||
+              hash(metadata)
```
matthiasS-da: MINOR: Previously, you had a version with a single code bIock covering all versions. I prefer that version, as it is more concise.

community/daml-lf/transaction/src/main/protobuf/com/digitalasset/daml/lf/transaction.proto
matthiasS-da: Thanks, all good.

community/common/src/test/scala/com/digitalasset/canton/protocol/hash/v2/NodeHashTest.scala
```diff
@@ -212,6 +212,16 @@ class NodeHashTest extends BaseNodeHashTest {
...
 
   "CreateNodeBuilder" should {
+    "suggest V3 for LF serialization V2 nodes" in {
+      val error = the[NodeHashingError.UnsupportedSerializationVersion] thrownBy hashCreateNode(
+        createNode.copy(version = SerializationVersion.V2)
+      )
+
+      error.getMessage should include(
+        "Please use hashing scheme V3 or higher."
+      )
```
matthiasS-da: MINOR: Not so important in this particular case, but as a general rule, could you please instruct claude to check the *full* error message in unit tests.

community/common/src/test/scala/com/digitalasset/canton/protocol/hash/v4/NodeHashTest.scala
```diff
@@ -0,0 +1,161 @@
...
+          |'66756e6374696f6e' # function (string)
+          |# Config
+          |'00000001' # 1 (int)
+          |'00' # config
+          |# Input
+          |'00000001' # 1 (int)
+          |'01' # input
+          |# Output
+          |'00000001' # 1 (int)
+          |'03' # output""".stripMargin
```
matthiasS-da, line 107: For this one, using "include" is fine from my point of view.

community/common/src/test/scala/com/digitalasset/canton/protocol/hash/v4/NodeHashTest.scala
```diff
@@ -0,0 +1,161 @@
...
+    "include exercise external-call result order and multiplicity" in {
+      val hash = hashExerciseNode(exerciseNode(ImmArray(externalCallResult1, externalCallResult2)))
+
+      hashExerciseNode(
+        exerciseNode(ImmArray(externalCallResult2, externalCallResult1))
+      ) should not be hash
+      hashExerciseNode(
+        exerciseNode(ImmArray(externalCallResult1, externalCallResult1))
+      ) should not be hash
+      hashExerciseNode(exerciseNode(ImmArray(externalCallResult1))) should not be hash
```
matthiasS-da: Wait, this does not check if the hash function is sensitive to multiplicity. You need to compare the last two left-hand sides with each other.

community/common/src/test/scala/com/digitalasset/canton/protocol/hash/v4/NodeHashTest.scala
```diff
@@ -0,0 +1,161 @@
...
+  "Pre-V4 NodeHashBuilder" should {
+    "reject external-call results rather than silently dropping them" in {
+      // External-call results must never be silently omitted from the hash: a scheme that does
+      // not encode them must fail when they are present. Each scheme is paired with the highest
+      // LF serialization version it accepts (V2 -> V1, V3 -> V2) so the failure is the
+      // external-call rejection, not the upfront version-compatibility check.
+      Seq(
+        HashingSchemeVersion.V2 -> SerializationVersion.V1,
+        HashingSchemeVersion.V3 -> SerializationVersion.V2,
+      ).foreach { case (schemeVersion, lfVersion) =>
```
matthiasS-da: Please rename lfVersion -> serializationVersion

matthiasS-da:
@angelol 

The branch currently does not compile in CI and I suggest to do a full recompile. For this, a file needs to be created on the branch. Unfortunately, I don't have permissions to do so.

Could you please either grant me permissions to push or apply the following patch?
```
Subject: [PATCH] Force a full recompilation
---
Index: .circleci/branches_to_be_fully_recompiled_in_ci.txt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/.circleci/branches_to_be_fully_recompiled_in_ci.txt b/.circleci/branches_to_be_fully_recompiled_in_ci.txt
new file mode 100644
--- /dev/null	(revision 2ca929ae38d739c31490555e563dfcff647a125f)
+++ b/.circleci/branches_to_be_fully_recompiled_in_ci.txt	(revision 2ca929ae38d739c31490555e563dfcff647a125f)
@@ -0,0 +1,5 @@
+^main$
+zenith-network:angelol/external-call-06-split-pr1-foundation
+fork/zenith-network/angelol/external-call-06-split-pr1-foundation
+zenith-network/angelol/external-call-06-split-pr1-foundation
+pull/549/head
```

matthiasS-da:
@angelol concerning test coverage: what are your plans of adding an integration test that exercises a prepare/execute request against the new hashing scheme version?

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
thibault-da: This is actually a doc forward port missing on main. The doc is up to date on 3.5 but not main. I've raised as a comment in the other PR and will make sure the missing doc gets forward ported

angelol:
> @angelol
> 
> The branch currently does not compile in CI and I suggest to do a full recompile. For this, a file needs to be created on the branch. Unfortunately, I don't have permissions to do so.
> 
> Could you please either grant me permissions to push or apply the following patch?
> 
> ```
> Subject: [PATCH] Force a full recompilation
> ---
> Index: .circleci/branches_to_be_fully_recompiled_in_ci.txt
> IDEA additional info:
> Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
> <+>UTF-8
> ===================================================================
> diff --git a/.circleci/branches_to_be_fully_recompiled_in_ci.txt b/.circleci/branches_to_be_fully_recompiled_in_ci.txt
> new file mode 100644
> --- /dev/null	(revision 2ca929ae38d739c31490555e563dfcff647a125f)
> +++ b/.circleci/branches_to_be_fully_recompiled_in_ci.txt	(revision 2ca929ae38d739c31490555e563dfcff647a125f)
> @@ -0,0 +1,5 @@
> +^main$
> +zenith-network:angelol/external-call-06-split-pr1-foundation
> +fork/zenith-network/angelol/external-call-06-split-pr1-foundation
> +zenith-network/angelol/external-call-06-split-pr1-foundation
> +pull/549/head
> ```

I have created that file

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
```diff
@@ -628,36 +643,76 @@ Create
 Exercise
 ^^^^^^^^
 
-.. code-block::
-
-    fn encode_node(exercise):
-        0x01 || # Node encoding version
```
matthiasS-da, line 634: This line has been removed and that seems to be correct.
@thibault-da could you please confirm?

matthiasS-da:
There is a test failing in CI:

https://app.circleci.com/pipelines/github/digital-asset/canton/180/workflows/de56d67c-538f-475c-b996-23b0d460b3b5/jobs/2814/tests

By now, you can remove the following lines from the file you have just created to speedup CI:
```
zenith-network:angelol/external-call-06-split-pr1-foundation
fork/zenith-network/angelol/external-call-06-split-pr1-foundation
zenith-network/angelol/external-call-06-split-pr1-foundation
pull/549/head
```

@angelol 
This concludes my review. Once the remaining findings are addressed, I'm happy to approve your PR!

community/common/src/test/scala/com/digitalasset/canton/protocol/hash/v4/NodeHashTest.scala
angelol: Fixed in 54da735da — the last two are now compared directly (`hash([r1, r1]) should not be hash([r1])`) so multiplicity is actually exercised; kept the order check against `[r1, r2]`.

community/common/src/test/scala/com/digitalasset/canton/protocol/hash/v4/NodeHashTest.scala
angelol: Renamed in 735145384.

community/common/src/test/scala/com/digitalasset/canton/protocol/hash/v2/NodeHashTest.scala
angelol: Switched to a full-message `shouldBe` in c0f3fb5e1, and I'll apply the full-message rule to error assertions going forward.

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
angelol: Restored the single code block in b4ce569f3 — it still notes both bytes (`0x03` for V3, `0x04` for V4), just without the tabs.

angelol:
Fixed in `0bf1e1ebd` — the assertions expected `Please using` where the message says `Please use`. Validated locally. Also removed `branches_to_be_fully_recompiled_in_ci.txt` in `4a6680b2a`.

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
thibault-da: Yes this is gone in V3 and V4. However V2 is still supported and we should keep this code block and put it into a `V2` tab

matthiasS-da: @angelol: I've talked to @thibault-da. We'll take care of it. No action needed.

matthiasS-da:
@angelol thanks for the fixes. From my point of view, the following points are outstanding:

1. An answer to [this question](https://github.com/digital-asset/canton/pull/549#issuecomment-4852377484).
2. Get CI green. (E.g., you still need to run `sbt format`. You can ignore the test_protocol_version_36 job, as the failed test is unrelated/flaky.)

angelol:
A full prepare/execute test with a real external-call choice needs Daml-language surface for the builtin, which today exists only at the LF level (`EXTERNAL_CALL`) — that is the companion PR digital-asset/daml#23099. Since #23099 can only merge once this Canton stack has landed (it depends on a published Canton artifact), the end-to-end test will follow as a subsequent step once the Daml side is published and pinned here.

matthiasS-da (commented):
LGTM - thanks for the effort and thanks for addressing all the comments!

matthiasS-da (approved):

angelol:
Format is fixed too — `f88ec82b6` (the v2 `NodeHashTest` assertion). With the test fix `0bf1e1ebd`, both the failing tests and the format check should come back green on a CI re-run.

matthiasS-da:
Post-merge cleanups:
* Add a prepare/execute integration test with hashing scheme version v4. ([blocked](https://github.com/digital-asset/canton/pull/549#issuecomment-4855538964)) (#564)
* Fix [documentation](https://github.com/digital-asset/canton/pull/549#discussion_r3505513485) of node encoding version in v2 vs v3+. (Thanks @thibault-da!)

docs-open/src/sphinx/sdk/explanations/external-signing/external_signing_hashing_algorithm.rst
matthiasS-da: To be fixed as part of [this PR](https://github.com/canton-network/cf-docs/pull/971). Thanks @thibault-da!

PR #551, opened 2026-06-30
feat(external-call): command-execution errors, rejection and handler SPI

angelol:
## Summary

This is PR 3 of 8 in the external-call runtime-integration series tracked in #513, stacked on PR 2.

The external-call stack (#513): #506 added the transaction-side representation for recording external-call results; #514 added the LF `EXTERNAL_CALL` builtin surface; #518 wired it into Speedy and the LF engine; #522 added transaction protobuf encoding/decoding; #526 added Canton protocol serialization for recorded results; #537 added the participant-side extension-service client. #541 implemented the remaining runtime integration as one PR; this series splits #541 into 8 independently reviewable PRs (each < 1000 lines) and supersedes it.

This PR adds the command-execution error surface and the handler service-provider interface (SPI) that later PRs implement and wire in.

## Scope

This PR adds:

- external-call-aware command-execution errors (`CommandExecutionErrors`) and the `ErrorResource` metadata for them
- rejection generators for external-call failures (`RejectionGenerators`)
- the external-call handler SPI (`ExternalCallHandler`, `ExternalCallMode`) — the interface used by command interpretation to dispatch engine external-call questions
- tests for rejection generation, request validation, and the LF/Speedy external-call surface

## Out of scope

The concrete extension-service handler implementation and its wiring into command execution (PR 5), engine replay (PR 4), and validation/routing/factory (PR 6–8). The SPI ships with a no-op `Unsupported` handler; nothing dispatches to a real service yet.

## Stacking & review

This PR is stacked on PR 2, so its diff against `main` is **cumulative**. The incremental change for this PR alone:
https://github.com/digital-asset/canton/compare/zenith-network:angelol/external-call-06-split-pr2-codec...zenith-network:angelol/external-call-06-split-pr3-interp

Refs #513.

PR #550, opened 2026-06-30
feat(external-call): prepared-transaction codec for recorded results

angelol:
## Summary

This is PR 2 of 8 in the external-call runtime-integration series tracked in #513, stacked on PR 1.

The external-call stack (#513): #506 added the transaction-side representation for recording external-call results; #514 added the LF `EXTERNAL_CALL` builtin surface; #518 wired it into Speedy and the LF engine; #522 added transaction protobuf encoding/decoding; #526 added Canton protocol serialization for recorded results; #537 added the participant-side extension-service client. #541 implemented the remaining runtime integration as one PR; this series splits #541 into 8 independently reviewable PRs (each < 1000 lines) and supersedes it.

This PR adds the interactive-submission representation and codec for recorded external-call results, so prepared transactions can round-trip them.

## Scope

This PR adds:

- the `ExternalCallResult` interactive-submission message (`interactive_submission_data.proto`) and the related service doc-comments (`interactive_submission_service.proto`)
- prepared-transaction encoding/decoding of recorded external-call results (`PreparedTransactionEncoder` / `PreparedTransactionDecoder`)
- the regenerated protobuf snapshot, `proto-data.yml` and `openapi.yaml` for the new message
- prepared-transaction codec tests

## Out of scope

Runtime dispatch of external-call questions (PR 3+), engine replay (PR 4), and validation (PR 6–8).

## Hashing Boundary

Prepared/external-submission signatures authorize the LF action. Recorded external-call result payloads are Canton protocol data protected by the Canton view signature (see PR 1). Prepared transactions therefore carry external-call results for round-tripping, but those result payloads are intentionally excluded from the LF/prepared node hash.

## Stacking & review

This PR is stacked on PR 1, so its diff against `main` is **cumulative** (it includes PR 1). The incremental change for this PR alone:
https://github.com/digital-asset/canton/compare/zenith-network:angelol/external-call-06-split-pr1-foundation...zenith-network:angelol/external-call-06-split-pr2-codec

Refs #513.

PR #552, opened 2026-06-30
feat(external-call): engine execution, model conformance and validation SPI

angelol:
## Summary

This is PR 4 of 8 in the external-call runtime-integration series tracked in #513, stacked on PR 3.

This PR adds the engine-side replay of recorded external-call results and the validation SPI used during confirmation.

## Scope

This PR adds:

- `DAMLe` reinterpretation support for replaying recorded external-call results, with `ExternalCallPayloadDescription`
- the `ExternalCallValidator` SPI trait and `ExternalCallValidationError`
- `ModelConformanceChecker` integration that replays recorded results during conformance checking
- engine, model-conformance and DAMLe tests

## Out of scope

The concrete validator implementation (PR 6), the extension-service runtime (PR 5), consistency checking and routing (PR 6–7), and activation of validation in transaction processing (PR 8).

## Stacking & review

PR 1–3 of the series have merged, and this branch is rebased onto `main` — the "Files changed" diff of this PR is now exactly its own incremental slice (no cumulative content from other PRs).

Refs #513.

### DIFF
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallValidator.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallValidator.scala
@@ -1,0 +1,23 @@
+// Copyright (c) 2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.digitalasset.canton.participant.protocol.validation
+
+import com.digitalasset.canton.lifecycle.FutureUnlessShutdown
+import com.digitalasset.canton.participant.util.DAMLe
+import com.digitalasset.canton.tracing.TraceContext
+import com.digitalasset.daml.lf.data.Bytes
+
+trait ExternalCallValidator {
+  def validate(
+      key: DAMLe.ExternalCallKey,
+      recordedOutput: Bytes,
+  )(implicit traceContext: TraceContext): FutureUnlessShutdown[ExternalCallValidator.Result]
+}
+
+object ExternalCallValidator {
+  sealed trait Result extends Product with Serializable
+  case object Matched extends Result
+  final case class Mismatched(computedOutput: Bytes, recordedOutput: Bytes) extends Result
+  final case class UnableToValidate(reason: String) extends Result
+}
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
@@ -274,32 +274,23 @@ class ModelConformanceChecker(
     })
 
   private def externalCallReplayDataFor(
-      view: TransactionView,
-      viewParticipantData: ViewParticipantData,
+      view: TransactionView
   )(implicit
       traceContext: TraceContext
   ): ExternalCallReplayData = {
-    val externalCallResults =
-      if (!viewParticipantData.supportsExternalCallResults)
-        Seq.empty[ViewParticipantData.ViewExternalCallResult]
-      else
-        view.flatten.flatMap { currentView =>
-          currentView.viewParticipantData.unwrap match {
-            case Right(vpd) => vpd.externalCallResults
-            case _ => Seq.empty
-          }
-        }
-
-    if (externalCallResults.isEmpty) ExternalCallReplayData.empty
-    else {
-      val replayData = ExternalCallReplayData.fromResults(externalCallResults.map(_.result))
-
-      logger.debug(
-        s"reInterpret: Aggregated ${replayData.size} external call result keys"
-      )
-
-      replayData
+    val externalCallResults = view.flatten.flatMap { currentView =>
+      currentView.viewParticipantData.unwrap match {
+        case Right(vpd) => vpd.externalCallResults
+        case _ => Seq.empty
+      }
     }
+    val replayData = ExternalCallReplayData.fromResults(externalCallResults.map(_.result))
+
+    logger.debug(
+      s"reInterpret: Aggregated ${replayData.size} external call result keys"
+    )
+
+    replayData
   }
 
   def reInterpret(
@@ -326,7 +317,7 @@ class ModelConformanceChecker(
     )
 
     lazy val externalCallReplayData: ExternalCallReplayData =
-      externalCallReplayDataFor(view, viewParticipantData)
+      externalCallReplayDataFor(view)
 
     for {
 
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
@@ -345,7 +345,7 @@ class ModelConformanceChecker(
           packagePreference,
           failed,
           getEngineAbortStatus,
-          () => FutureUnlessShutdown.pure(externalCallReplayData),
+          () => externalCallReplayData,
         )(traceContext)
         .leftMap(DAMLeError(_, view.viewHash))
         .leftWiden[Error]
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
@@ -292,14 +292,13 @@ class ModelConformanceChecker(
 
     if (externalCallResults.isEmpty) ExternalCallReplayData.empty
     else {
-      val storedExternalCallResults =
-        StoredExternalCallResults.fromResults(externalCallResults.map(_.result))
+      val replayData = ExternalCallReplayData.fromResults(externalCallResults.map(_.result))
 
       logger.debug(
-        s"reInterpret: Aggregated ${storedExternalCallResults.size} external call result keys"
+        s"reInterpret: Aggregated ${replayData.size} external call result keys"
       )
 
-      ExternalCallReplayData(storedExternalCallResults)
+      replayData
     }
   }
 
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
@@ -273,6 +273,36 @@ class ModelConformanceChecker(
       } yield nameBindings
     })
 
+  private def externalCallReplayDataFor(
+      view: TransactionView,
+      viewParticipantData: ViewParticipantData,
+  )(implicit
+      traceContext: TraceContext
+  ): ExternalCallReplayData = {
+    val externalCallResults =
+      if (!viewParticipantData.supportsExternalCallResults)
+        Seq.empty[ViewParticipantData.ViewExternalCallResult]
+      else
+        view.flatten.flatMap { currentView =>
+          currentView.viewParticipantData.unwrap match {
+            case Right(vpd) => vpd.externalCallResults
+            case _ => Seq.empty
+          }
+        }
+
+    if (externalCallResults.isEmpty) ExternalCallReplayData.empty
+    else {
+      val storedExternalCallResults =
+        StoredExternalCallResults.fromResults(externalCallResults.map(_.result))
+
+      logger.debug(
+        s"reInterpret: Aggregated ${storedExternalCallResults.size} external call result keys"
+      )
+
+      ExternalCallReplayData(storedExternalCallResults)
+    }
+  }
+
   def reInterpret(
       view: TransactionView,
       ledgerTime: CantonTimestamp,
@@ -296,6 +326,9 @@ class ModelConformanceChecker(
       view.viewParticipantData.tryUnwrap.keyResolution.fmap(_.unversioned.contracts),
     )
 
+    lazy val externalCallReplayData: ExternalCallReplayData =
+      externalCallReplayDataFor(view, viewParticipantData)
+
     for {
 
       packagePreference <- buildPackageNameMap(packageIdPreference, topologySnapshot, ledgerTime)
@@ -313,6 +346,7 @@ class ModelConformanceChecker(
           packagePreference,
           failed,
           getEngineAbortStatus,
+          () => FutureUnlessShutdown.pure(externalCallReplayData),
         )(traceContext)
         .leftMap(DAMLeError(_, view.viewHash))
         .leftWiden[Error]
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
@@ -199,10 +199,10 @@ object DAMLe {
       )
   }
 
-  /** Stored external call results for replay during reinterpretation. Multiple outputs for one
+  /** External-call replay data: recorded outputs indexed by semantic key. Multiple outputs for one
     * semantic key are preserved so replay can report a recorded-result disagreement.
     */
-  final case class StoredExternalCallResults private (
+  final case class ExternalCallReplayData private (
       outputsByKey: Map[ExternalCallKey, Set[LfBytes]]
   ) {
     def size: Int = outputsByKey.size
@@ -217,26 +217,17 @@ object DAMLe {
       }
   }
 
-  object StoredExternalCallResults {
-    val empty: StoredExternalCallResults = StoredExternalCallResults(Map.empty)
+  object ExternalCallReplayData {
+    val empty: ExternalCallReplayData = ExternalCallReplayData(Map.empty)
 
-    def fromResults(results: Iterable[ExternalCallResult]): StoredExternalCallResults =
-      StoredExternalCallResults(
+    def fromResults(results: Iterable[ExternalCallResult]): ExternalCallReplayData =
+      ExternalCallReplayData(
         results
           .groupMap(ExternalCallKey.fromResult)(_.output)
           .view
           .mapValues(_.toSet)
           .toMap
       )
-  }
-
-  final case class ExternalCallReplayData(
-      storedExternalCallResults: StoredExternalCallResults
-  )
-
-  object ExternalCallReplayData {
-    val empty: ExternalCallReplayData =
-      ExternalCallReplayData(StoredExternalCallResults.empty)
   }
 
   trait HasReinterpret {
@@ -484,7 +475,7 @@ class DAMLe(
       }
 
       replayDataF.flatMap { replayData =>
-        replayData.storedExternalCallResults.outputFor(externalCallKey) match {
+        replayData.outputFor(externalCallKey) match {
           case Left(disagreement) =>
             failExternalCall(disagreement)
 
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
@@ -437,11 +437,8 @@ class DAMLe(
     def handleExternalCall(
         externalCallKey: ExternalCallKey,
         resume: Either[ResultNeedExternalCall.Error, String] => Result[A],
-        replayDataO: Option[ExternalCallReplayData],
-    ): FutureUnlessShutdown[Either[ReinterpretationError, A]] = {
-      val replayData = replayDataO.getOrElse(externalCallReplayData())
-
-      replayData.outputFor(externalCallKey) match {
+    ): FutureUnlessShutdown[Either[ReinterpretationError, A]] =
+      externalCallReplayData().outputFor(externalCallKey) match {
         case Left(disagreement) =>
           failExternalCall(disagreement)
 
@@ -452,13 +449,11 @@ class DAMLe(
           logger.debug(
             s"Replaying recorded external call result for extension=${externalCallKey.extensionId}, function=${externalCallKey.functionId}"
           )
-          handleResultInternal(resume(Right(storedOutput.toHexString)), Some(replayData))
+          handleResultInternal(resume(Right(storedOutput.toHexString)))
       }
-    }
 
     def handleResultInternal(
-        result: Result[A],
-        replayDataO: Option[ExternalCallReplayData],
+        result: Result[A]
     ): FutureUnlessShutdown[Either[ReinterpretationError, A]] = {
 
       @tailrec
@@ -486,7 +481,7 @@ class DAMLe(
             )
             .transformWithHandledAborted {
               case Success(pkg) =>
-                handleResultInternal(resume(pkg), replayDataO)
+                handleResultInternal(resume(pkg))
               case Failure(ex) =>
                 logger.error(s"Package resolution failed for [$packageId]", ex)
                 FutureUnlessShutdown.failed(ex)
@@ -518,10 +513,7 @@ class DAMLe(
                     ResultNeedKey.Response.UnsupportedContractIdVersion(contract.contractId)
                 }
               }
-              handleResultInternal(
-                resume(ResultNeedKey.Response(entries, hasStarted)),
-                replayDataO,
-              )
+              handleResultInternal(resume(ResultNeedKey.Response(entries, hasStarted)))
             }
 
         case ResultNeedContract(acoid, resume) =>
@@ -542,7 +534,7 @@ class DAMLe(
             }
           FutureUnlessShutdown
             .pure(response)
-            .flatMap(r => handleResultInternal(resume(r), replayDataO))
+            .flatMap(r => handleResultInternal(resume(r)))
 
         case ResultError(err) => FutureUnlessShutdown.pure(Left(EngineError(err)))
         case ResultInterruption(continue, _) =>
@@ -550,11 +542,11 @@ class DAMLe(
           // Using a `Future` as a trampoline also makes the recursive call to `handleResult` stack safe.
           FutureUnlessShutdown.pure(iterateOverInterrupts(continue)).flatMap {
             case Left(abort) => FutureUnlessShutdown.pure(Left(abort))
-            case Right(result) => handleResultInternal(result, replayDataO)
+            case Right(result) => handleResultInternal(result)
           }
         case ResultPrefetch(_, _, resume) =>
           // we do not need to prefetch here as Canton includes the keys as a static map in Phase 3
-          handleResultInternal(resume(), replayDataO)
+          handleResultInternal(resume())
         case ResultNeedExternalCall(extensionId, functionId, configHash, input, resume) =>
           val externalCallKey = ExternalCallKey(
             extensionId,
@@ -562,15 +554,11 @@ class DAMLe(
             configHash,
             input,
           )
-          handleExternalCall(
-            externalCallKey,
-            resume,
-            replayDataO,
-          )
+          handleExternalCall(externalCallKey, resume)
       }
     }
 
-    handleResultInternal(result, None)
+    handleResultInternal(result)
   }
 
 }
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
@@ -14,6 +14,11 @@ import com.digitalasset.canton.logging.{NamedLoggerFactory, NamedLogging}
 import com.digitalasset.canton.participant.protocol.EngineController.GetEngineAbortStatus
 import com.digitalasset.canton.participant.store.ReplayContractLookup
 import com.digitalasset.canton.participant.util.DAMLe.*
+import com.digitalasset.canton.participant.util.ExternalCallPayloadDescription.{
+  byteCount,
+  byteSize,
+  hexPayloadSize,
+}
 import com.digitalasset.canton.protocol.*
 import com.digitalasset.canton.topology.ParticipantId
 import com.digitalasset.canton.topology.client.TopologySnapshot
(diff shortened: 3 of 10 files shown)


community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceCheckerTest.scala
paulbrauner-da, line 1: Is there anything else to this file change than inlining? If not then I'd suggest reverting it for readability of the PR.

community/participant/src/test/scala/com/digitalasset/canton/participant/util/DAMLeExternalCallTest.scala
```diff
@@ -0,0 +1,203 @@
...
+      inside(runReinterpret()) { case Right(result) =>
+        val exerciseNodes = result.transaction.nodes.collect { case (_, exercise: Node.Exercise) =>
+          exercise
+        }
+        exerciseNodes should have size 1
+        exerciseNodes.head.externalCallResults shouldBe ImmArray(externalCallResult)
+      }
+    }
+
+    "reject duplicate semantic outputs without choosing one" in {
```
paulbrauner-da: did you mean "conflicting"?

community/participant/src/test/scala/com/digitalasset/canton/participant/util/DAMLeExternalCallTest.scala
```diff
@@ -0,0 +1,203 @@
...
+
+      inside(
+        runReinterpret(
+          storedExternalCallResultsForReplay = storedResults
+        )
+      ) { case Left(disagreement: DAMLe.ExternalCallRecordedResultDisagreement) =>
+        disagreement.toString should not include externalCallResult.output.toHexString
+        disagreement.toString should not include conflictingOutput.output.toHexString
+        disagreement.toString should not include externalCallResult.config.toHexString
+        disagreement.toString should not include externalCallResult.input.toHexString
```
paulbrauner-da, line 181: I'm not clear on what this is testing: the test case description says "without choosing one", so you might think that this proves that it didn't pick one of the calls for the error message. But rather it seems to show that we don't leak either of the inputs and ouputs in the case of a conflicting response.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
```diff
@@ -120,6 +129,116 @@ object DAMLe {
...
+  }
+
+  final case class ExternalCallReplayData(
+      storedExternalCallResults: StoredExternalCallResults
+  )
+
+  object ExternalCallReplayData {
+    val empty: ExternalCallReplayData =
+      ExternalCallReplayData(StoredExternalCallResults.empty)
+  }
```
paulbrauner-da: Could you explain why we have this wrapping for `StoredExternalCallResults`? Or said otherwise: why can't `StoredExternalCallResults` be renamed `ExternalCallReplayData` and delete `ExternalCallReplayData`?

paulbrauner-da (commented):
First batch of (harmless) comments. I have possibly more consequential comments queued, but I need to discuss them internally first.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
```diff
@@ -133,6 +252,7 @@ object DAMLe {
         packageResolution: Map[Ref.PackageName, Ref.PackageId],
         expectFailure: Boolean,
         getEngineAbortStatus: GetEngineAbortStatus,
+        externalCallReplayData: () => FutureUnlessShutdown[ExternalCallReplayData],
```
paulbrauner-da: What's the reason this needs to return a `FutureUnlessShutdown`? Does the caller typically retrieve this replay data from an async source?

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceCheckerTest.scala
angelol: Nothing else, just inlining. Reverted in 4797f861a to keep the PR focused.

community/participant/src/test/scala/com/digitalasset/canton/participant/util/DAMLeExternalCallTest.scala
angelol: They're conflicting — two different outputs for one input identity, not duplicate. Renamed in 6f88615f2.

community/participant/src/test/scala/com/digitalasset/canton/participant/util/DAMLeExternalCallTest.scala
angelol: Agreed, the name misled here. The behaviour: external-call outputs are deterministic per input, so two *different* recorded outputs for one input is a violation that must be rejected, not resolved by picking one. The test asserts both — the result is a `Left(Disagreement)` (rejected, not chosen), and its message doesn't leak the payloads. Renamed to match the assertions in 6f88615f2; the missing-output non-leakage case is the separate test just below.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
angelol: Agreed — no reason for the wrapper. Merged `StoredExternalCallResults` into `ExternalCallReplayData` (kept that name) and deleted the redundant type in dde7f095b.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
angelol: No — `ModelConformanceChecker` builds it synchronously from an in-memory `lazy val`, so there's no async source. Dropped the `FutureUnlessShutdown`; the thunk is now `() => ExternalCallReplayData` in dc20282f1.

matthiasS-da:
Starting reviewing...

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallValidator.scala
```diff
@@ -0,0 +1,23 @@
...
+import com.digitalasset.canton.participant.util.DAMLe
+import com.digitalasset.canton.tracing.TraceContext
+import com.digitalasset.daml.lf.data.Bytes
+
+trait ExternalCallValidator {
+  def validate(
+      key: DAMLe.ExternalCallKey,
+      recordedOutput: Bytes,
+  )(implicit traceContext: TraceContext): FutureUnlessShutdown[ExternalCallValidator.Result]
+}
```
matthiasS-da, line 16: What's the reason of adding this code?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
```diff
@@ -273,6 +273,35 @@ class ModelConformanceChecker(
...
+        Seq.empty[ViewParticipantData.ViewExternalCallResult]
+      else
+        view.flatten.flatMap { currentView =>
+          currentView.viewParticipantData.unwrap match {
+            case Right(vpd) => vpd.externalCallResults
+            case _ => Seq.empty
+          }
+        }
+
+    if (externalCallResults.isEmpty) ExternalCallReplayData.empty
```
matthiasS-da: What is the benefit of this additional if?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
```diff
@@ -273,6 +273,35 @@ class ModelConformanceChecker(
...
     })
 
+  private def externalCallReplayDataFor(
+      view: TransactionView,
+      viewParticipantData: ViewParticipantData,
+  )(implicit
+      traceContext: TraceContext
+  ): ExternalCallReplayData = {
+    val externalCallResults =
+      if (!viewParticipantData.supportsExternalCallResults)
```
matthiasS-da: What is the benefit of this if? If external call results are unsupported, the else branch evaluates to an empty Seq as well.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
```diff
@@ -120,6 +129,107 @@ object DAMLe {
...
+      param("extension id", _.key.extensionId.doubleQuoted),
+      param("function id", _.key.functionId.doubleQuoted),
+      param(
+        "config bytes",
+        disagreement => hexPayloadSize(disagreement.key.config).doubleQuoted,
+      ),
+      param(
+        "input bytes",
+        disagreement => hexPayloadSize(disagreement.key.input).doubleQuoted,
+      ),
```
matthiasS-da: Please simplify this (also for the next class below) as follows:
1. Have ExternalCallKey implement PrettyPrinting.
2. Simply add `param("key", _.key)`.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
```diff
@@ -120,6 +129,107 @@ object DAMLe {
...
+  ) {
+    def size: Int = outputsByKey.size
+
+    def outputFor(
+        key: ExternalCallKey
+    ): Either[ExternalCallRecordedResultDisagreement, Option[LfBytes]] =
+      outputsByKey.get(key) match {
+        case None => Right(None)
+        case Some(outputs) if outputs.sizeCompare(1) == 0 => Right(outputs.headOption)
+        case Some(outputs) => Left(ExternalCallRecordedResultDisagreement(key, outputs))
```
matthiasS-da, line 205: This is a very important validation step, but from my point of view the implementation is unnecessarily complex, as it is spread over ViewParticipantData, ModelConformanceChecker and DAMLe.
Have you considered putting this into TransactionView?
- Fail TransactionView.validated if there are differing outputs for the same key.
- Add a method to TransactionView that computes the ExternalCallReplayData.
- Change ExternalCallReplayDate.outputsByKey to have the type Map[ExternalCallKey, LfBytes].

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
```diff
@@ -120,6 +129,107 @@ object DAMLe {
...
+        "input bytes",
+        missing => hexPayloadSize(missing.key.input).doubleQuoted,
+      ),
+    )
+  }
+
+  /** Deterministic external-call identity. Config and input are engine-emitted canonical hex
+    * strings.
+    */
+  final case class ExternalCallKey(
```
matthiasS-da, line 162: This is an optional design improvement for code cleanliness and is not required for approval. Feel free to track this as a post-merge refactoring chore.

From an architectural standpoint, it looks like ExternalCallKey was introduced primarily to avoid typing out Map[(String, String, String, String), Set[LfBytes]] inside ExternalCallReplayData.

If that's the case, we could improve encapsulation in one of two ways:
1. Move ExternalCallKey directly into object ExternalCallReplayData as a private class (which would require a quick adjustment to the signature of ExternalCallReplayData.outputFor).
2. Alternatively, update ResultNeedExternalCall to take a parameter of type ExternalCallKey.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
```diff
@@ -323,12 +436,61 @@ class DAMLe(
...
+            resume,
+            Some(replayData),
+          )
+      }
+    }
 
     def handleResultInternal(
-        result: Result[A]
+        result: Result[A],
+        replayDataO: Option[ExternalCallReplayData],
```
matthiasS-da: It seems that replayDataO is always None. If that's the case, please remove the parameter.

matthiasS-da:
@angelol to make the compilation pass, please force full recompilation as in #549.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
```diff
@@ -323,12 +436,61 @@ class DAMLe(
...
+      FutureUnlessShutdown.pure(Left(error))
+
+    def resumeWithExternalCallOutput(
+        output: String,
+        resume: Either[ResultNeedExternalCall.Error, String] => Result[A],
+        replayDataO: Option[ExternalCallReplayData],
+    ): FutureUnlessShutdown[Either[ReinterpretationError, A]] =
+      handleResultInternal(resume(Right(output)), replayDataO)
+
+    def replayStoredExternalCallOutput(
```
matthiasS-da: There seems to be only a single usage - please inline this method for simplicity.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
```diff
@@ -323,12 +436,61 @@ class DAMLe(
...
+      externalCallReplayData: () => ExternalCallReplayData,
   )(implicit
       traceContext: TraceContext
   ): FutureUnlessShutdown[Either[ReinterpretationError, A]] = {
+    def failExternalCall(
+        error: ReinterpretationError
+    ): FutureUnlessShutdown[Either[ReinterpretationError, A]] =
+      FutureUnlessShutdown.pure(Left(error))
+
+    def resumeWithExternalCallOutput(
```
matthiasS-da: There seems to be only a single usage - please inline this method for simplicity.

matthiasS-da:
@angelol from my point of view the following tests are needed:
1. End-to-end test with external calls
2. Test with disagreeing external call outputs (different views have different outputs for the same key)
3. Test with missing external call outputs (The engine request the output for a key. The key is not contained by any external call result in the request.)

For 1, I understand this is blocked as long as you don't have external calls in Daml.
What is your plan for 2 and 3?

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ExampleTransactionConformanceTest.scala
```diff
@@ -444,16 +450,162 @@ class ExampleTransactionConformanceTest
...
+              reconstructed.viewParticipantData.tryUnwrap.externalCallResults.loneElement
 
             receivedRecord.checkingParties shouldBe Set.empty
             reconstructedRecord.checkingParties shouldBe Set(submitter)
           }
         }
       }
 
+      "aggregate visible external-call results for replay" in {
+        val devFactory = new ExampleTransactionFactory(versionOverride = Some(ProtocolVersion.dev))(
```
matthiasS-da: Please do not hardcode protocol versions in tests. The reason is that we run every test with every supported protocol version. By hardcoding the version, the same test runs several times.

Instead:
- Use testedProtocolVersion to get the currently tested protocol version.
- Change this test case to `"aggregate visible external-call results for replay" onlyRunWithOrGreaterThan(ProtocolVersion.dev) in { ` so that it runs only with protocol version dev.
- Have another test case running only with protocol version < dev, testing the desired behavior (e.g. empty replay data) for these protocol versions.

matthiasS-da:
I've concluded my first review pass:

- Please address my detailed comments above.
- If the compile job keeps failing, please apply the same [trick](https://github.com/digital-asset/canton/pull/549#issuecomment-4852473843) as in #549.
- Please rebase this PR as soon as #550 and #551 have been merged to main.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
angelol: You're right, there is none — `ViewParticipantData.validated` already rejects non-empty `externalCallResults` below dev on every construction path, so the guard was dead. Removed in 0c94bc7fa9.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
angelol: None — `fromResults(empty)` is already `empty`. Removed in 0c94bc7fa9, same commit as the guard below; the method is now a plain aggregate-and-log.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
angelol: Done in 5b26633db9. `ExternalCallKey` prints payload sizes rather than values so the error messages keep not leaking payloads; as a bonus this also shields the key's own default `toString`, which did contain the raw hex.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
angelol: Inlined in be4f24eda7 (together with the other single-use helper below — they were one chain).

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
angelol: Inlined in be4f24eda7.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
angelol: Almost — it was `Some` on the resume path: it memoized the thunk result across multiple external calls within one reinterpretation. But the only production caller already passes a memoized `lazy val`, so the internal caching was redundant. Removed the parameter and the threading in a4be0ee0ab; `handleExternalCall` now calls the thunk directly.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ExampleTransactionConformanceTest.scala
angelol: Done in 6125756c44. The test now uses the suite factory + `testedProtocolVersion`, gated `onlyRunWithOrGreaterThan(ProtocolVersion.dev)` like the tampered-metadata test above, and there's a counterpart below dev asserting the aggregated replay data stays empty.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallValidator.scala
angelol: It's the SPI for validating a recorded external-call result against the extension service — same situation as the handler SPI in #551: the implementation arrives with the extension service (#553) and the consumers with the consistency checker and response router (#554/#555). It rides in this PR with the other external-call validation types.

angelol:
Applied the same full-recompilation trick as #549 in b4a740b839, and addressed the detailed comments above (one commit each).

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
angelol: The question never posed itself, because this code was written against the full feature rather than this slice — and in the full series the validation you're pointing at deliberately lives elsewhere. `ExternalCallConsistencyChecker` (#554) detects conflicting recorded outputs structurally across all visible occurrences — including the occurrences and their checking parties — and `ExternalCallResponseRouter` (#555) turns exactly that into per-checking-party consistency rejections plus an alarm; #556 suppresses the replay-side duplicate so it isn't double-reported. Failing `TransactionView.validated` instead would reject the whole tree as malformed, dropping the per-party confirmation semantics and the occurrence granularity a key-level check can't carry. The check here is intentionally narrow: a self-contained guard that reinterpretation never proceeds on ambiguous inputs — which is also why `outputsByKey` keeps the `Set`; it's what lets replay detect the conflict without depending on the checker having run. So I'd keep this as is — happy to revisit after #554/#555 if you still see a better cut.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
angelol: The key stays public because the consistency checker and response router later in the series (#554/#555) group and order by it, which rules out option 1; option 2 would change the engine's `ResultNeedExternalCall` API, which lives on the Daml side. Happy to track a post-merge tidy-up once the full series is in.

angelol:
On 2 and 3:

**3 (missing output)** is covered in this PR: `DAMLeExternalCallTest` "reject external calls without recorded output" has the engine request a key with no recorded result and asserts the `ExternalCallReplayMissing` rejection; "not leak external-call payloads for replay errors" covers the same path's error hygiene.

**2 (disagreeing outputs)** has its replay-level half in this PR — "reject conflicting semantic outputs without leaking their payloads" (two recorded outputs for one key fail reinterpretation) — but the cross-view semantics you're describing are deliberately Phase-3 validation and are tested where they land later in the series: the consistency checker in #554 covers detection directly (e.g. "report only hosted parties that check conflicting outputs", "record visible disagreements without reporting non-hosted checking parties", "not report identical outputs"), and #555/#556 cover how those disagreements become confirmation responses (e.g. "route recorded external-call result disagreements by checking party", "split external-call disagreements from the general verdict by party", "alarm on visible recorded external-call disagreements without hosted confirmers").

**For 1**, the plan is an end-to-end prepare/execute test as soon as daml#23099 gives external calls a Daml-language surface — there's nothing to hang it on before that.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
matthiasS-da: Ok, thanks for the explanation. We can defer the discussion until we reach the reviews for #554 and #555. Please note that I am uneasy with this approach; relying on logic in future unmerged PRs to justify current complexity is highly prone to causing architectural friction and unexpected delays, especially under such a tight timeline.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
```diff
@@ -323,9 +425,32 @@ class DAMLe(
...
       result: Result[A],
       getEngineAbortStatus: GetEngineAbortStatus,
+      externalCallReplayData: () => ExternalCallReplayData,
   )(implicit
       traceContext: TraceContext
   ): FutureUnlessShutdown[Either[ReinterpretationError, A]] = {
+    def failExternalCall(
+        error: ReinterpretationError
+    ): FutureUnlessShutdown[Either[ReinterpretationError, A]] =
+      FutureUnlessShutdown.pure(Left(error))
```
matthiasS-da: MINOR: With the current state, I'd inline this as well for conciseness.

matthiasS-da:
> On 2 and 3:
> 
> **3 (missing output)** is covered in this PR: `DAMLeExternalCallTest` "reject external calls without recorded output" has the engine request a key with no recorded result and asserts the `ExternalCallReplayMissing` rejection; "not leak external-call payloads for replay errors" covers the same path's error hygiene.
> 
> **2 (disagreeing outputs)** has its replay-level half in this PR — "reject conflicting semantic outputs without leaking their payloads" (two recorded outputs for one key fail reinterpretation) — but the cross-view semantics you're describing are deliberately Phase-3 validation and are tested where they land later in the series: the consistency checker in #554 covers detection directly (e.g. "report only hosted parties that check conflicting outputs", "record visible disagreements without reporting non-hosted checking parties", "not report identical outputs"), and #555/#556 cover how those disagreements become confirmation responses (e.g. "route recorded external-call result disagreements by checking party", "split external-call disagreements from the general verdict by party", "alarm on visible recorded external-call disagreements without hosted confirmers").
> 
> **For 1**, the plan is an end-to-end prepare/execute test as soon as daml#23099 gives external calls a Daml-language surface — there's nothing to hang it on before that.

Thanks for explaining. I'm aware of the unit tests in DAMLeExternalCallTest. I intended to refer to end-to-end integration tests, where a malicious submitting participant produces these situations. What's the plan for that?

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
angelol: Understood — though one clarification that may ease this: the later PRs aren't future design work. The full feature exists as one implementation (this series is #541 sliced for reviewability), so the checker and router I'm referring to are finished code in #554/#555, open for inspection today.

angelol:
For 2 and 3: both need the confirming participant to reinterpret a transaction that genuinely contains external-call nodes — a malicious submitter doesn't change that. Without a real external-call node behind the recorded results, the honest side rejects on model conformance rather than the external-call path, so the disagreement / missing-output logic never runs. So these need an external-call choice in a Daml package, same as the honest e2e (1).

Two options:
1. Land them with the external-call Daml surface (daml#23099), against a compiled package.
2. Now, without it: the LF builtin exists in dev-LF, so I could hand-author a raw-LF package with an external-call choice (in-memory via the LF test parser) and exercise it through the ledger API. Caveats: I found no precedent for exercising a hand-authored package at the integration level (execution tests all use compiled DARs), it'd be pinned to dev-LF, and presumably replaced once (1) lands.

The harness is otherwise ready, so either way it's a small delta. Interim coverage is at the component level: `ExternalCallConsistencyCheckerTest`, `TransactionConfirmationResponsesFactoryExternalCallTest`, `DAMLeExternalCallTest`.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
angelol: Inlined in 3f92db572e.

matthiasS-da:
> For 2 and 3: both need the confirming participant to reinterpret a transaction that genuinely contains external-call nodes — a malicious submitter doesn't change that. Without a real external-call node behind the recorded results, the honest side rejects on model conformance rather than the external-call path, so the disagreement / missing-output logic never runs. So these need an external-call choice in a Daml package, same as the honest e2e (1).
> 
> Two options:
> 
> 1. Land them with the external-call Daml surface (daml#23099), against a compiled package.
> 2. Now, without it: the LF builtin exists in dev-LF, so I could hand-author a raw-LF package with an external-call choice (in-memory via the LF test parser) and exercise it through the ledger API. Caveats: I found no precedent for exercising a hand-authored package at the integration level (execution tests all use compiled DARs), it'd be pinned to dev-LF, and presumably replaced once (1) lands.
> 
> The harness is otherwise ready, so either way it's a small delta. Interim coverage is at the component level: `ExternalCallConsistencyCheckerTest`, `TransactionConfirmationResponsesFactoryExternalCallTest`, `DAMLeExternalCallTest`.

Ok, let's defer tests 1-3 to a later PR.

matthiasS-da:
@angelol #551 has been merged. Please rebase this one!

paulbrauner-da:
> Caveats: I found no precedent for exercising a hand-authored package at the integration level (execution tests all use compiled DARs), it'd be pinned to dev-LF, and presumably replaced once (1) lands.

We actually have such an integration test. The dars are produced [here](https://github.com/DACH-NY/canton/blob/18e710b8fd60d1b2a2c0f8ef8b658332e93f0841/community/daml-lf/upgrades-matrix-integration/src/test/scala/com/digitalasset/canton/integration/tests/UpgradesMatrixIT.scala#L150) from [this LF code](https://github.com/DACH-NY/canton/blob/18e710b8fd60d1b2a2c0f8ef8b658332e93f0841/community/daml-lf/upgrades-matrix/src/main/scala/com/digitalasset/daml/lf/engine/UpgradesMatrix.scala#L453). So it's not unheard of.

angelol:
Rebased onto latest main (post-#551), along with the rest of the stack — the diff is now the clean PR-4 slice. Local gates are green (whole-repo test-compile + the dev-PV suites for this PR's surface).

angelol:
Thanks — I'd missed UpgradesMatrixIT, that removes the no-precedent caveat. When we pick up the deferred tests, producing the DAR from hand-authored LF per that pattern looks like the right route.

angelol:
Following up on https://github.com/digital-asset/canton/pull/551#discussion_r3518217233 (#551 is locked): the force-recompile file is removed here in 8d61347 — this PR's diff is now the 8 content files only. The equivalent files on PR 5–8 will be dropped as each is rebased.

matthiasS-da:
Resuming review...

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
matthiasS-da: Please move this class to a separate file in com.digitalasset.canton.participant.protocol.validation.

Happy to track that as post-merge cleanup task.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
matthiasS-da: Thanks again for the detailed breakdown, @angelol. Having now looked ahead and reviewed the entire trajectory of this design across PRs 5–8, I am still strongly in favor of centralizing this in TransactionView rather than scattering it across ModelConformanceChecker and DAMLe.

However, to avoid deadlocking your current review train, I am willing to defer this structural change to an explicit, post-merge architectural cleanup task.

For the record, and to frame that upcoming cleanup, here are the core technical reasons why the current layout introduces structural regressions:

- Performance Bottleneck: It does not make sense to spend expensive cycles running transaction reinterpretation on a view that is guaranteed to be discarded anyway.

- Fail-Last Complexity: Carrying an invalid state forward that is destined to fail introduces significant accidental complexity downstream. The need for PR #556 to explicitly suppress duplicate rejections is a direct symptom of this "fail-last" design pattern, and reviewing this extra code to handle the split responsibility inherently consumes more review cycles as we go down the chain.

Regarding the specific downstream trade-offs you mentioned:

> Failing TransactionView.validated instead would reject the whole tree as malformed, ...

Note that the protocol will ultimately reject the whole tree (recte: whole view) anyway, even with the downstream machinery introduced in PRs 5–8.

> ...dropping the per-party confirmation semantics and the occurrence granularity a key-level check can't carry.

I don't think the per-party confirmation semantics provide any practical advantage in this scenario. Every hosted confirming party can observe the disagreement; therefore, it is entirely sufficient to reject on behalf of every hosted confirming party rather than building a complex routing mechanism to isolate them.

Let's proceed with this slice as-is for now so you can move forward, on the strict condition that we open a tracked technical debt issue to refactor this validation logic back to a centralized, fail-fast model post-milestone.

matthiasS-da (approved):
Approving with the caveats outlined in the detailed comments. 

Most importantly: 

1. The additional complexity of processing views with result disagreements should be removed.
2. Missing end-to-end tests need to be implement.

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
matthiasS-da: See #565

community/participant/src/main/scala/com/digitalasset/canton/participant/util/DAMLe.scala
matthiasS-da: See #565

matthiasS-da:
> @angelol from my point of view the following tests are needed:
> 
> 1. End-to-end test with external calls
> 2. Test with disagreeing external call outputs (different views have different outputs for the same key)
> 3. Test with missing external call outputs (The engine request the output for a key. The key is not contained by any external call result in the request.)
> 
> For 1, I understand this is blocked as long as you don't have external calls in Daml. What is your plan for 2 and 3?

Missing end-to-end tests are tracked in #564.

PR #555, opened 2026-06-30
feat(external-call): confirmation-response router

angelol:
## Summary

This is PR 7 of 8 in the external-call runtime-integration series tracked in #513, stacked on PR 6.

The external-call stack (#513): #506 added the transaction-side representation for recording external-call results; #514 added the LF `EXTERNAL_CALL` builtin surface; #518 wired it into Speedy and the LF engine; #522 added transaction protobuf encoding/decoding; #526 added Canton protocol serialization for recorded results; #537 added the participant-side extension-service client. #541 implemented the remaining runtime integration as one PR; this series splits #541 into 8 independently reviewable PRs (each < 1000 lines) and supersedes it.

This PR adds the router that turns recorded-result consistency outcomes into confirmation-response routing decisions, directing disagreements to the relevant checking parties.

## Scope

This PR adds:

- `ExternalCallResponseRouter`, which maps consistency-check outcomes to confirmation-response routing (including routing disagreements to the relevant checking parties), plus the relocated `ViewWithHostedParties` router input
- a focused `ExternalCallResponseRouterTest` covering the router's scenarios directly against its API

## Out of scope

Assembly of confirmation responses and activation of validation in transaction processing (PR 8).

## Stacking & review

This PR is stacked on PR 6, so its diff against `main` is **cumulative**. The incremental change for this PR alone:
https://github.com/digital-asset/canton/compare/zenith-network:angelol/external-call-06-split-pr6-checker...zenith-network:angelol/external-call-06-split-pr7-router

Refs #513.

PR #556, opened 2026-06-30
feat(external-call): confirmation-responses factory and validation activation

angelol:
## Summary

This is the concluding PR (8 of 8) in the external-call runtime-integration series tracked in #513, stacked on PR 7.

This PR runs external-call validation as one of the parallel validation suites of transaction processing and translates its outcome into confirmation responses — completing the Canton-side runtime integration.

## Scope

This PR adds:

- the external-call validation suite: `ExternalCallResponseRouter.check` is invoked from `doParallelChecks` in `TransactionProcessingSteps`, following the `ModelConformanceChecker.check` pattern (kicked off there, awaited when responses are created). It resolves the hosted confirming parties, checks recorded-result consistency (`ExternalCallConsistencyChecker`), alarms on visible disagreements, re-validates undisputed recorded results against the extension service concurrently with the reinterpretation, and routes replay disagreements chained on the model-conformance errors. Requests without recorded external-call results short-circuit without topology lookups. The precomputed result travels through `TransactionValidationResult`.
- external-call handling in `TransactionConfirmationResponsesFactory`, which stays a pure translation layer: one nested def splits a view's verdict per party from the precomputed result (disagreement rejections, re-validation rejections, abstentions, remainder verdict), with rejections taking precedence per party and view.
- the validator wiring that activates external-call validation in transaction processing (`TransactionProcessor` → `TransactionProcessingSteps`, `ConnectedSynchronizer`, `CantonSyncService`, `SynchronizerConnectionsManager`, and the `ParticipantNode` validator binding)
- tests: the router check-suite tests and the confirmation-responses factory translation tests

## Out of scope

Nothing further in the Canton-side stack — this PR completes it. The feature remains `ProtocolVersion.dev`-gated and is inert unless an extension service is configured.

## Coordination

This series supersedes #541. #537 is merged; merge this series in order PR 1 → PR 8 (each rebased onto `main` as the previous lands).

Companion Daml PR: https://github.com/digital-asset/daml/pull/23099. It can remain open for parallel review, but after this Canton series lands it must be amended to pin a normal Canton snapshot/artifact containing these runtime constants and error contracts, with Daml verification rerun against that published artifact before it merges.

## Stacking & review

This PR is stacked on PR 7, so its diff against `main` is **cumulative** — it shows the whole feature. The incremental change for this PR alone:
https://github.com/digital-asset/canton/compare/zenith-network:angelol/external-call-06-split-pr7-router...zenith-network:angelol/external-call-06-split-pr8-factory

Note: rebased onto main HEAD (post-#552); the temporary cherry-picks of PR 6 / PR 7 review fixes are gone (their content now reaches this branch through the stack).

Refs #513.

### DIFF
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessor.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessor.scala
@@ -16,7 +16,6 @@ import com.digitalasset.base.error.{
 }
 import com.digitalasset.canton.*
 import com.digitalasset.canton.concurrent.FutureSupervisor
-import com.digitalasset.canton.config.RequireTypes.PositiveInt
 import com.digitalasset.canton.config.{ProcessingTimeout, TestingConfigInternal}
 import com.digitalasset.canton.crypto.*
 import com.digitalasset.canton.data.CantonTimestamp
@@ -42,6 +41,7 @@ import com.digitalasset.canton.participant.protocol.submission.{
 }
 import com.digitalasset.canton.participant.protocol.validation.{
   AuthorizationValidator,
+  ExternalCallResponseRouter,
   ExternalCallValidator,
   InternalConsistencyChecker,
   ModelConformanceChecker,
@@ -104,8 +104,6 @@ class TransactionProcessor(
         new TransactionConfirmationResponsesFactory(
           participantId,
           synchronizerId,
-          externalCallValidator,
-          TransactionProcessor.externalCallValidationParallelism(participantNodeParameters),
           loggerFactory,
         ),
         ModelConformanceChecker(
@@ -130,6 +128,12 @@ class TransactionProcessor(
         InternalConsistencyChecker(
           participantId,
           staticSynchronizerParameters.protocolVersion,
+          loggerFactory,
+        ),
+        new ExternalCallResponseRouter(
+          participantId,
+          externalCallValidator,
+          participantNodeParameters.general.batchingConfig.parallelism,
           loggerFactory,
         ),
         commandProgressTracker,
@@ -217,11 +221,6 @@ class TransactionProcessor(
 }
 
 object TransactionProcessor {
-
-  private def externalCallValidationParallelism(
-      participantNodeParameters: ParticipantNodeParameters
-  ): PositiveInt =
-    participantNodeParameters.general.batchingConfig.parallelism
 
   sealed trait TransactionProcessorError
       extends WrapsProcessorError
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessor.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessor.scala
@@ -16,6 +16,7 @@ import com.digitalasset.base.error.{
 }
 import com.digitalasset.canton.*
 import com.digitalasset.canton.concurrent.FutureSupervisor
+import com.digitalasset.canton.config.RequireTypes.PositiveInt
 import com.digitalasset.canton.config.{ProcessingTimeout, TestingConfigInternal}
 import com.digitalasset.canton.crypto.*
 import com.digitalasset.canton.data.CantonTimestamp
@@ -41,6 +42,7 @@ import com.digitalasset.canton.participant.protocol.submission.{
 }
 import com.digitalasset.canton.participant.protocol.validation.{
   AuthorizationValidator,
+  ExternalCallValidator,
   InternalConsistencyChecker,
   ModelConformanceChecker,
   TransactionConfirmationResponsesFactory,
@@ -87,6 +89,7 @@ class TransactionProcessor(
     promiseFactory: PromiseUnlessShutdownFactory,
     participantNodeParameters: ParticipantNodeParameters,
     trafficEnforcementBackendO: Option[TrafficEnforcementBackend],
+    externalCallValidator: ExternalCallValidator,
 )(implicit val ec: ExecutionContext)
     extends ProtocolProcessor[
       TransactionProcessingSteps.SubmissionParam,
@@ -101,6 +104,8 @@ class TransactionProcessor(
         new TransactionConfirmationResponsesFactory(
           participantId,
           synchronizerId,
+          externalCallValidator,
+          TransactionProcessor.externalCallValidationParallelism(participantNodeParameters),
           loggerFactory,
         ),
         ModelConformanceChecker(
@@ -212,6 +217,11 @@ class TransactionProcessor(
 }
 
 object TransactionProcessor {
+
+  private def externalCallValidationParallelism(
+      participantNodeParameters: ParticipantNodeParameters
+  ): PositiveInt =
+    participantNodeParameters.general.batchingConfig.parallelism
 
   sealed trait TransactionProcessorError
       extends WrapsProcessorError
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
@@ -134,9 +134,10 @@ class TransactionConfirmationResponsesFactory(
     /** Converts the local verdict of a view into confirmation responses, splitting it by party
       * where precomputed external-call outcomes ([[ExternalCallResponseRouter.Result]]) apply:
       * hosted confirming parties that check a disagreeing external-call result reject the view with
-      * the disagreement, parties whose checked results could not be re-validated abstain, and the
-      * remaining parties respond with the view's own verdict. For a request without external calls,
-      * `externalCallResult` is empty and the view's verdict is returned unsplit.
+      * the disagreement, parties whose checked results could not be re-validated abstain (on views
+      * that would otherwise be approved), and the remaining parties respond with the view's own
+      * verdict. For a request without external calls, `externalCallResult` is empty and the view's
+      * verdict is returned unsplit.
       */
     def responsesForVerdict(
         viewPosition: ViewPosition,
@@ -256,8 +257,11 @@ class TransactionConfirmationResponsesFactory(
               // An external-call replay disagreement that is routed to a hosted checking party is
               // not turned into (nor logged as) a malformed model-conformance rejection of the
               // whole view: the routed parties instead reject the affected views with a dedicated,
-              // equally logged ExternalCallResultDisagreement rejection (see responsesForVerdict),
-              // so every model-conformance error still surfaces in exactly one logged rejection.
+              // equally logged ExternalCallResultDisagreement rejection (see responsesForVerdict).
+              // A routed error is never silent: its visible-disagreement alarm has fired in the
+              // external-call check, and each affected view logs its rejection when emitting it --
+              // unless the view is rejected as malformed for another reason, in which case that
+              // stronger rejection wins the view (the alarm remains on record).
               Option.unless(externalCallResult.isRoutableModelConformanceError(cause))(
                 logged(
                   requestId,
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
@@ -131,14 +131,13 @@ class TransactionConfirmationResponsesFactory(
       }
     }
 
-    /** Converts the local verdict of a view into confirmation responses, splitting it by party
-      * where precomputed external-call outcomes ([[ExternalCallResponseRouter.Result]]) apply:
-      * hosted confirming parties that check a disagreeing external-call result reject the view with
-      * the disagreement, parties whose checked results could not be re-validated abstain (on views
-      * that would otherwise be approved), and the remaining parties respond with the view's own
-      * verdict. For a request without external calls, `externalCallResult` is empty and the view's
-      * verdict is returned unsplit.
-      */
+    // Converts the local verdict of a view into confirmation responses, splitting it by party
+    // where precomputed external-call outcomes (ExternalCallResponseRouter.Result) apply:
+    // hosted confirming parties that check a disagreeing external-call result reject the view
+    // with the disagreement, parties whose checked results could not be re-validated abstain (on
+    // views that would otherwise be approved), and the remaining parties respond with the view's
+    // own verdict. For a request without external calls, `externalCallResult` is empty and the
+    // view's verdict is returned unsplit.
     def responsesForVerdict(
         viewPosition: ViewPosition,
         hostedConfirmingParties: Set[LfPartyId],
@@ -214,9 +213,10 @@ class TransactionConfirmationResponsesFactory(
                 checked(
                   ConfirmationResponse.tryCreate(
                     Some(viewPosition),
-                    LocalAbstainError.CannotPerformAllValidations
-                      .Abstain(reason)
-                      .toLocalAbstain(protocolVersion),
+                    logged(
+                      requestId,
+                      LocalAbstainError.CannotPerformAllValidations.Abstain(reason),
+                    ).toLocalAbstain(protocolVersion),
                     parties.toSet,
                   )
                 )
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
@@ -4,7 +4,6 @@
 package com.digitalasset.canton.participant.protocol.validation
 
 import cats.syntax.parallel.*
-import com.digitalasset.canton.config.RequireTypes.PositiveInt
 import com.digitalasset.canton.data.{CantonTimestamp, ViewPosition}
 import com.digitalasset.canton.error.TransactionError
 import com.digitalasset.canton.lifecycle.FutureUnlessShutdown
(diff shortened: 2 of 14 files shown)


matthiasS-da:
Start reviewing...

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessor.scala
```diff
@@ -212,6 +217,11 @@ class TransactionProcessor(
 
 object TransactionProcessor {
 
+  private def externalCallValidationParallelism(
```
matthiasS-da: MINOR: Please inline.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -23,12 +24,251 @@ import scala.concurrent.ExecutionContext
...
+      externalCallValidationParallelism,
+      loggerFactory,
+    )
+
+  private def responseForView(
+      viewPosition: ViewPosition,
+      localVerdict: LocalVerdict,
+      parties: Set[LfPartyId],
+  ): ConfirmationResponse =
+    checked(
```
matthiasS-da: The checked variable indicates that the preconditions for tryCreate (called two lines below) have been successfully met. In the previous iteration of the code, this safety invariant was completely obvious due to the local structure of the logic. With this new layout, that relationship is now obscured.

To ensure long-term maintainability and safe local reasoning, I strongly recommend reverting to the original structure where this precondition check was explicit and adjacent.

At a minimum, please refactor this block to a structure that makes it visually obvious why tryCreate is guaranteed never to throw an exception here, and back it up with a clear code comment explaining the invariant. We shouldn't be leaving potential runtime panics hidden behind implicit code positioning.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -23,12 +24,251 @@ import scala.concurrent.ExecutionContext
...
+          externalCallAbstainResponses ++
+          Option
+            .when(remainingParties.nonEmpty)(
+              responseForView(viewPosition, localVerdict, remainingParties)
+            )
+            .toList
+    }
+  }
+
+  private def ordinaryLocalVerdictAndParties(
```
matthiasS-da: Could you please undo this method extraction? Moving this code right now completely obfuscates the actual logical diff we need to review, and it dramatically increases the risk of nasty merge conflicts during the final push to main.

Let’s keep the code exactly where it was historically for this PR series so we can safely track the actual changes. If this refactor is genuinely needed for cleanliness, let's track it as an isolated chore to be done after the milestone lands.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -23,12 +24,251 @@ import scala.concurrent.ExecutionContext
...
+    checked(
+      ConfirmationResponse
+        .tryCreate(
+          Some(viewPosition),
+          localVerdict,
+          parties,
+        )
+    )
+
+  private def responsesForView(
```
matthiasS-da: I’m trying to follow the architectural convention for how functionality is being decomposed into methods here, but the separation of concerns feels a bit tangled.

We already have ordinaryLocalVerdictAndParties, which seems to be the designated place for computing verdicts from validation results. However, the verdicts based on external call validation are being computed here instead of there. On top of that, this method is pulling double-duty by doing generic verdict handling (e.g., mapping None to Seq.empty) and converting those results into final confirmation responses.

Could you please share the underlying motivation behind splitting the verdict logic across these different layers? If there isn’t a strict architectural reason for this layout, I strongly recommend reverting to the previous design. It will make it much easier to review the changes.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -133,151 +373,98 @@ class TransactionConfirmationResponsesFactory(
...
-                      LocalRejectError.MalformedRejects.MalformedRequest.Reject(
-                        err.format(viewPosition)
-                      ),
-                    ).toLocalReject(protocolVersion)
-                  )
+        // Rejections due to a failed model conformance check
+        // Aborts are logged by the Engine callback when the abort happens
+        modelConformanceRejections =
+          modelConformanceErrors.flatMap { case cause =>
+            Option.unless(externalCallRouting.isRoutableModelConformanceError(cause))(
```
matthiasS-da: This new condition silently switches off the logging of model conformance errors under certain circumstances. What is the explicit motivation for this?

Our standard convention is to log all validation errors within this method for visibility and auditability. Introducing silent paths where errors disappear without an explicitly documented reason makes production troubleshooting incredibly difficult. If this suppression is absolutely necessary for the protocol, please back it up with a clear inline comment explaining the invariant; otherwise, this condition should be removed.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -133,151 +373,98 @@ class TransactionConfirmationResponsesFactory(
...
-              val authorizationRejections =
-                transactionValidationResult.authorizationResult
-                  .get(viewPosition)
-                  .map(cause =>
-                    logged(
-                      requestId,
-                      LocalRejectError.MalformedRejects.MalformedRequest.Reject(cause),
-                    ).toLocalReject(protocolVersion)
-                  )
+        ordinaryViewResults = viewsWithHostedParties.map { viewWithHostedParties =>
```
matthiasS-da: I noticed that the view processing logic here has been changed from parallel (parTraverseFilter) to sequential (map). What is the underlying reason for dropping parallel processing?

Switching to sequential evaluation represents a significant architectural regression that could easily introduce latency bottlenecks and degrade transaction validation throughput under load. If there is a strict concurrency or state synchronization dependency that forces this to be sequential, please document it explicitly with a code comment. Otherwise, this should be reverted back to a parallel execution pattern to preserve our performance baselines.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -133,151 +373,98 @@ class TransactionConfirmationResponsesFactory(
...
-                      // ensures that this rejection or something at least as strong will make it to the mediator.
-                      LocalRejectError.MalformedRejects.MalformedRequest.Reject(
-                        err.format(viewPosition)
-                      ),
-                    ).toLocalReject(protocolVersion)
-                  )
+        // Rejections due to a failed model conformance check
+        // Aborts are logged by the Engine callback when the abort happens
+        modelConformanceRejections =
+          modelConformanceErrors.flatMap { case cause =>
```
matthiasS-da: MINOR: superfluous "case" keyword.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -133,151 +373,98 @@ class TransactionConfirmationResponsesFactory(
...
-              val contractConsistencyRejections =
-                transactionValidationResult.contractConsistencyResultE.swap.toOption.map(err =>
-                  logged(
-                    requestId,
-                    LocalRejectError.MalformedRejects.MalformedRequest.Reject(err.toString),
-                  ).toLocalReject(protocolVersion)
-                )
+        approvingViewPositions =
+          ordinaryViewResults.collect {
+            case (viewWithHostedParties, Some((_: LocalApprove, parties))) if parties.nonEmpty =>
```
matthiasS-da: MINOR: What is the purpose of the if? For LocalApprove, parties should never be empty.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -133,151 +373,98 @@ class TransactionConfirmationResponsesFactory(
...
-                  .get(viewPosition)
-                  .map(err =>
-                    logged(
-                      requestId,
-                      LocalRejectError.MalformedRejects.MalformedRequest.Reject(err.message),
-                    ).toLocalReject(protocolVersion)
-                  )
+        _ = externalCallRouter.reportVisibleRecordedDisagreementAlarms(
+          requestId,
+          externalCallRouting.recordedConsistencyResult,
```
matthiasS-da: Under the hood, this logic invokes ExternalCallConsistencyChecker. Beyond the fact that the name recordedConsistencyResult is misleading (it implies an $O(1)$ property retrieval when it is actually computing heavy validation from scratch), this breaks the foundational separation of concerns for this component.The sole purpose of TransactionConfirmationResponseFactory is to act as a pure translation layer: it should take existing validation results, report any errors, and convert them into confirmation responses. It should absolutely not be orchestrating or executing its own independent validation pipelines. The correct place to run this validation is within doParallelChecks inside TransactionProcessingSteps. Please move the invocation of ExternalCallConsistencyChecker there, and restrict this factory component to its designated role of simply mapping the checker's output into confirmation responses.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -133,151 +373,98 @@ class TransactionConfirmationResponsesFactory(
...
-                  case malformed: LocalReject if malformed.isMalformed => malformed -> Set.empty
-                  case localReject: LocalReject if hostedConfirmingParties.nonEmpty =>
-                    localReject -> hostedConfirmingParties
-                }
-                .orElse(
-                  Option.when(hostedConfirmingParties.nonEmpty)(
-                    LocalApprove(protocolVersion) -> hostedConfirmingParties
-                  )
-                )
+        localValidationRoutes <- externalCallRouter.validateExternalCalls(
```
matthiasS-da: Again, validateExternalCalls is executing heavy validation logic here, including external HTTPS network calls to outside services.

As outlined above, this is entirely the wrong layer to orchestrate validation, but it also carries severe performance implications: by invoking it here, this heavy network I/O is forced to run sequentially after all other validations have already completed, rather than executing concurrently alongside them. This creates an artificial synchronization bottleneck in the transaction pipeline. Please move this invocation to doParallelChecks within TransactionProcessingSteps so that it runs in parallel with the other validation suites (matching the pattern established by the invocation of ModelConformanceChecker.check). The factory layer should strictly be restricted to processing the resulting payload.

matthiasS-da:
I am pausing my deep, line-by-line review of this PR. The current implementation introduces significant architectural fragmentation, leaky abstractions, and performance problems that cannot be resolved via localized patches or quick fixes.

Before this PR can be re-evaluated, it requires a systematic structural revision to address four foundational design defects:

- **Leaky Factory Abstractions:** TransactionConfirmationResponseFactory must be restricted to its designated role as a pure translation layer (mapping validation results to confirmation responses). It must not orchestrate validation pipelines or execute heavy calculations under the hood via ExternalCallConsistencyChecker or ExternalCallValidator.

- **Sequential Performance Bottlenecks:** Heavy network I/O (validateExternalCalls) and view processing logic must not be forced into sequential execution paths (map) after the main validation loop. These must be moved to doParallelChecks in TransactionProcessingSteps to run concurrently alongside our other validation suites.

- **Tangled Method Decomposition:** The methods computing verdicts and confirmation responses have become severely entangled. I strongly recommend going back to the previous method structure. If changes are needed, keep methods focused on a single responsibility and avoid overlapping responsibilities between methods.

- **Obscured Fulfillment of Pre-conditions:** The code must be organized such that it is obvious that method pre-conditions (such as of tryCreate) are met.

- **Git Hygiene:** Unrelated cosmetic method extractions that obfuscate the actual logical diff of the protocol changes must be reverted to ensure history remains clear and to eliminate preventable merge conflicts.

Once the PR has been completely reorganized to respect these architectural boundaries, please request another pass. Because this initial review was cut short by these global structural issues, I reserve the right to point out further logical, edge-case, or implementation problems in subsequent iterations once the foundational architecture is stable.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: Done in 667edff19c. ExternalCallConsistencyChecker now runs inside
`ExternalCallResponseRouter.check`, invoked from doParallelChecks; the factory receives its
outcome as precomputed data (`ExternalCallResponseRouter.Result`) and only maps it into
responses. The misleading name is gone together with the hidden computation.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: Done in 667edff19c, following the ModelConformanceChecker.check pattern: `check(...)` is
kicked off in doParallelChecks and starts the re-validation network I/O before the
reinterpretation result is available, so it runs concurrently with the other validation
suites; the factory only awaits and folds the outcome. Two consequences of the concurrent
kickoff, both documented at `validateExternalCalls`: (1) at kickoff time it is not yet known
which views will be approved, so undisputed recorded results are re-validated regardless
(deduplicated per semantic key, parallelism-bounded); the factory still folds re-validation
outcomes only into approving views. (2) A key whose recorded outputs disagree among the
occurrences selected for re-validation (those checked by hosted, not-already-rejecting
confirming parties) is not re-validated, now independently of the per-view verdicts --
previously that depended on which views ended up approving. The disagreement itself is alarmed, and hosted
parties that check conflicting occurrences reject via the consistency check; a new router
test locks this behavior for the disjoint-checking-parties case.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: Reverted to the previous design in 667edff19c: the per-view verdict computation
(localVerdicts / collectFirst) is back inline in responsesForWellFormedPayloads, unchanged
from main. The one genuinely new step is party-splitting: external-call verdicts are per
(view, party) rather than per view -- a disagreement must not reject the whole view for
parties that do not check the result -- so a view's verdict can fan out into several
responses. That fan-out now lives in a single nested def (responsesForVerdict) whose only
job is converting the verdict plus the precomputed external-call outcomes into responses.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: Done in 667edff19c; the extraction is reverted and the rejection blocks are byte-identical to
main again. The factory's diff vs main is now +129/-17 (from +453/-135).

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: Done in 667edff19c. The original structure is restored, and each of the four tryCreate call
sites carries a comment stating why the invariant holds at that site (malformed verdicts come
with empty parties by construction of localVerdictAndPartiesO; groupMap groups are non-empty;
remainingParties is guarded by nonEmpty).

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: It is rerouting rather than suppression, and it is now documented inline as you suggested
(667edff19c, comment at the modelConformanceRejections guard): a replay disagreement that is
routed to at least one hosted checking party is rejected -- and logged -- as a per-party
ExternalCallResultDisagreement rejection for the affected views, instead of a view-global
malformed ModelConformance rejection. The feature's core semantics is that a disagreement
must not malform the view for the parties that do not check the result. The guard is true
only if the error actually routed to a hosted party, and a routed error is never silent: its
visible-disagreement alarm has fired in the external-call check, and each affected view logs
the dedicated rejection when emitting it -- unless a view is rejected as malformed for another
reason, in which case that stronger rejection wins the view with the alarm remaining on
record.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: Restored in 667edff19c: per-view processing is again a single parallel traversal
(parFlatTraverse, since a view can now yield several party-split responses). For the record,
the previous revision had kept the topology lookups -- the only effect in this loop --
parallel in a pre-pass, and the flagged map was pure; the pre-pass existed only because the
inline validation needed the cross-view approval set first, and it disappeared together with
that dependency.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessor.scala
angelol: Done in 667edff19c; the helper is gone and the expression is inlined at its single use site,
the ExternalCallResponseRouter construction.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: Gone in 667edff19c (the surrounding code was restructured; the new closure does not use
`case`).

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: You are right that LocalApprove always comes with non-empty parties. The collect and its
guard are gone in 667edff19c: the approval set is no longer computed at all, since
re-validation is not gated on it anymore.

angelol:
The PR has been restructured along the lines you set out. All commits are appended (no
force-push); 667edff19c carries the structural rework:

- **Factory as a pure translation layer.** TransactionConfirmationResponsesFactory is back to
  its pre-series shape (constructor, inline per-view verdict computation, single parallel
  per-view traversal, malformed-payloads branch). It runs no validation: it consumes a
  precomputed `ExternalCallResponseRouter.Result` and maps it into responses. Its diff vs main
  is down to the external-call additions (+129/-17, from +453/-135): one nested def
  (`responsesForVerdict`) that splits a view's verdict per party, and the model-conformance
  rerouting guard, now with an inline comment stating the invariant.
- **Validation moved to doParallelChecks.** `ExternalCallResponseRouter.check(...)` is invoked
  from doParallelChecks following the ModelConformanceChecker.check pattern: kicked off there,
  awaited when responses are created. It resolves the hosted confirming parties, runs the
  ExternalCallConsistencyChecker, alarms on visible disagreements, starts the re-validation
  network I/O before the reinterpretation result is available (so the two run concurrently),
  and routes replay disagreements chained on the conformance errors. Requests without recorded
  external-call results short-circuit to an empty result without topology lookups.
- **Two consequences worth calling out:** re-validation can no longer be gated on which views
  would otherwise approve (approval is not known at kickoff time; the factory still folds
  re-validation outcomes only into approving views, and rejections take precedence per party
  and view). First, undisputed results may be re-validated needlessly for views that end up
  rejected: extra calls against the participant's own extension service, deduplicated per
  semantic key and parallelism-bounded -- the same trade the model conformance check makes.
  Second, a key whose recorded outputs
  disagree among the occurrences selected for re-validation (those checked by hosted,
  not-already-rejecting confirming parties) is not re-validated, now independently of the
  per-view verdicts (documented at `validateExternalCalls`, locked by a new router test).
  Previously this skip depended on unrelated per-view verdicts: with all conflicting views
  approving, such a key was already skipped -- the disagreement is alarmed, and any hosted
  party that checks conflicting occurrences rejects via the consistency check -- and only when
  one conflicting view happened to fail some other check was the remaining side re-validated.
- **Possible sharper semantics for the disagreeing-key corner:** per-(key, output)
  re-validation -- validate each distinct recorded output independently, so parties checking a
  wrong output reject while parties checking the correct one approve. This is stronger than
  the pre-rework behavior as well (which also skipped such keys whenever both sides were
  selected), removes the single-output special case in `validateExternalCalls`, and costs at
  most one extra validator call per disagreeing key. Happy to implement it on this PR if you
  think it is the right semantics; otherwise I would propose it as a post-merge follow-up.
- **Method decomposition / tryCreate.** Verdict computation is back where it was historically;
  every ConfirmationResponse.tryCreate call site carries its invariant justification inline.
- **Git hygiene.** The cosmetic extractions are reverted.

Also included: the branch is synced with the current review state of PRs 6 and 7
(cherry-picks; they drop out at the post-merge rebases), and the CI force-recompile file is
removed (62e6a1b457), as on PR 4.

Each inline comment has an in-thread reply. Ready for another pass.

angelol:
Following up on the #552 approval conditions (its threads are locked post-merge): the fail-fast centralization of disagreement validation in TransactionView is now tracked as #565 — including the downstream machinery in this PR that it will obsolete (the per-party disagreement routing and the routable-error suppression), and the ExternalCallKey file move. The missing end-to-end and negative-path tests from the #552/#553 approval caveats are tracked as #564.

matthiasS-da:
@angelol ExternalCallResponseRouter.scala looks quite different than in #555. Could you please rebase this PR on top of #555, so I can peek into it to understand how exactly ExternalCallResponseRouter will be used? That will speed up my review.

angelol:
Done — rebased onto #555's current tip (48318ba22d); new tip d7012b3739, mergeable. The incremental slice is the compare link in the description; with the branches synced it now shows exactly this PR's changes on top of PR 7.

On the router difference you noticed: that is the restructure from your review pause. In #555's revision the factory drives the router's pieces inline; this PR turns it into a validation suite — `ExternalCallResponseRouter.check(...)` is invoked from `doParallelChecks` (the ModelConformanceChecker pattern), and the factory only translates the precomputed `ExternalCallResponseRouter.Result` into per-view, per-party responses. The quickest orientation is the class ScalaDoc's "Intended data flow" paragraph and `check`'s method doc; the former `ExternalCallRoutingContext` is absorbed into `Result`.

matthiasS-da:
Unsubscribing for now as this has been superseded by #567. Happy to resume if you request it.

Issue #565, opened 2026-07-06
External-call: centralize disagreement validation in TransactionView (fail-fast) + review cleanup items


angelol:
Tracking issue for the post-merge follow-ups agreed during the external-call runtime-integration reviews (#549–#556, tracked in #513). Part 1 is the architectural condition attached to the #552 approval; Part 2 collects the smaller cleanup items the reviews explicitly deferred.

## Part 1 — Centralize disagreement validation in `TransactionView` (fail-fast)

Condition recorded at [#552 review discussion](https://github.com/digital-asset/canton/pull/552#discussion_r3506482637): refactor the external-call recorded-result disagreement handling from the current fail-last model to a centralized, fail-fast model in `TransactionView`, post-milestone.

Reviewer rationale (summarizing the discussion):
- A view carrying a recorded-result disagreement is guaranteed to be rejected; reinterpreting it through the engine spends cycles on a view destined to be discarded.
- Carrying the invalid state forward creates accidental complexity downstream — the per-party disagreement routing and the suppression of duplicate model-conformance rejections in the confirmation-response path are direct symptoms.
- Per-party confirmation granularity provides no practical advantage for this scenario: every hosted confirming party can observe the disagreement, so rejecting on behalf of all of them suffices.

Expected shape of the refactor:
- [x] Detect recorded-result disagreements at `TransactionView` validation and fail the view as malformed there.
- [x] Remove the downstream machinery this obsoletes: replay-disagreement surfacing through model-conformance errors (`DAMLe`/`ModelConformanceChecker`), the per-party disagreement routing, and the routable-error suppression in the confirmation-response path.
- [x] Re-examine the re-validation occurrence selection afterwards: with disagreements rejected fail-fast, keys with disagreeing recorded outputs no longer reach re-validation, which also settles the selected-occurrence corner discussed at [#556](https://github.com/digital-asset/canton/pull/556#issuecomment-4879132690) (and supersedes the per-(key, output) re-validation alternative offered there).

## Part 2 — Review cleanup items

From #552:
- [x] Move `DAMLe.ExternalCallKey` to a separate file in `com.digitalasset.canton.participant.protocol.validation` ([discussion](https://github.com/digital-asset/canton/pull/552#discussion_r3506626583)).

From #553 (all marked deferrable at approval):
[...]

Issue #564, opened 2026-07-06
External-call: end-to-end and negative-path integration tests (post-merge follow-up for #549-#556)


angelol:
Tracking issue for the end-to-end and negative-path test work agreed during the external-call runtime-integration reviews (#549–#556, tracked in #513). All items were explicitly deferred to post-merge by reviewer agreement.

### End-to-end integration tests (deferred from #552, [comment](https://github.com/digital-asset/canton/pull/552#issuecomment-4873979099))
- [x] Honest-path e2e: prepare/execute a transaction with a real external-call choice against a (mock) extension service, through the full confirmation workflow.
- [x] Disagreeing outputs: recorded external-call results that disagree, asserting the resulting rejections and the visible-disagreement alarm.
- [x] Missing output: recorded replay data absent for an executed call.
- [x] Prepare/execute IT with hashing scheme V4 (deferred from #549, [comment](https://github.com/digital-asset/canton/pull/549#issuecomment-4856349487)).

Feasibility note: possible without digital-asset/daml#23099 by hand-authoring a dev-LF package with the textual-LF parser and uploading it, following the `UpgradesMatrixIT` precedent (pointer in [#552 discussion](https://github.com/digital-asset/canton/pull/552#issuecomment-4877213770)). The mock-extension-service wiring is the one unprecedented piece.

### Negative-path tests (requested at #553 approval, [comment](https://github.com/digital-asset/canton/pull/553#issuecomment-4891915028))
- [x] Failure cases of `ExternalCallHandler`: assert both the failures reported to ledger-api clients and the emitted log messages. (Unit level in #591, ledger-api level in #634; the `Unsupported` handler is unit-covered only — an integration test would need a participant with no extensions configured, deliberately skipped for marginal value.)
- [x] Failure cases of `HttpExtensionServiceClient` (status-code mapping, retries, non-canonical output), with the same client-visible + log assertions.

Refs #513.

angelol:
Two more post-merge test cleanups, agreed during the #567 review:

- [x] Do not hardcode the protocol version in `ExternalCallProtocolIntegrationTest` — derive it via the usual dev-gating instead ([#567 comment](https://github.com/digital-asset/canton/pull/567#discussion_r3535114765)).
[...]

PR #554, opened 2026-06-30
feat(external-call): recorded-result consistency checker and extension-service validator

angelol:
## Summary

This is PR 6 of 8 in the external-call runtime-integration series tracked in #513, stacked on PR 5.

The external-call stack (#513): #506 added the transaction-side representation for recording external-call results; #514 added the LF `EXTERNAL_CALL` builtin surface; #518 wired it into Speedy and the LF engine; #522 added transaction protobuf encoding/decoding; #526 added Canton protocol serialization for recorded results; #537 added the participant-side extension-service client. #541 implemented the remaining runtime integration as one PR; this series splits #541 into 8 independently reviewable PRs (each < 1000 lines) and supersedes it.

This PR adds the consistency checker that compares recorded external-call results against locally checked results, and the concrete extension-service validator implementing the PR 4 SPI.

## Scope

This PR adds:

- `ExternalCallConsistencyChecker`, which compares recorded results with locally checked results and identifies disagreements
- `ExtensionServiceExternalCallValidator`, the extension-service implementation of the PR 4 `ExternalCallValidator` SPI
- consistency-checker tests and the shared external-call validation test fixtures used here and in PR 7

## Out of scope

Routing of disagreements to checking parties (PR 7) and activation of the validator in transaction processing (PR 8). The validator is implemented here but not yet threaded into confirmation; it is dev-gated and dormant until PR 8.

## Stacking & review

This PR is stacked on PR 5, so its diff against `main` is **cumulative**. The incremental change for this PR alone (it depends on PR 3 and PR 4, not PR 5):
https://github.com/digital-asset/canton/compare/zenith-network:angelol/external-call-06-split-pr5-extension...zenith-network:angelol/external-call-06-split-pr6-checker

Refs #513.

PR #567, opened 2026-07-07
feat(external-call): minimal protocol integration

angelol:
## Summary

This PR completes the minimal protocol integration of external calls (PR "6b" of the series tracked in #513, as agreed with @matthiasS-da): recorded external-call results are validated during transaction confirmation, and any disagreement rejects the request on behalf of all hosted confirming parties.

The external-call stack (#513): #506 added the transaction-side representation for recording external-call results; #514 added the LF `EXTERNAL_CALL` builtin surface; #518 wired it into Speedy and the LF engine; #522 added transaction protobuf encoding/decoding; #526 added Canton protocol serialization for recorded results; #537 added the participant-side extension-service client; #549–#554 added hashing (V4), the prepared-transaction codec, errors and the handler SPI, engine execution and the validation SPI, extension-service wiring, and the consistency checker. This PR adds the missing runtime consumer.

## Scope

- **`ExternalCallCheck`**: a validation suite invoked from `doParallelChecks` in `TransactionProcessingSteps`, following the `ModelConformanceChecker.check` pattern (kicked off there, awaited when confirmation responses are created, network I/O concurrent with the other suites). It checks consistency of the recorded results across the request (`ExternalCallConsistencyChecker`; every visible disagreement is alarmed) and re-validates undisputed recorded outputs against the configured extension service, once per distinct call.
- **Verdicts** (via the response factory's existing per-view verdict chain — its structure is unchanged): a disagreement from either part rejects each view on behalf of all its hosted confirming parties (`LOCAL_VERDICT_EXTERNAL_CALL_RESULT_DISAGREEMENT`); a recorded result that cannot be re-validated (e.g. no extension service configured) downgrades approvals to `CANNOT_PERFORM_ALL_VALIDATIONS` abstentions. One consequence of keeping the factory's first-match chain: the new non-malformed rejection ranks with the other consistency rejections, so — like the pre-existing time and contract-consistency rejections — it is reported in preference to a co-occurring malformed rejection from the later suites (authentication, model conformance, internal consistency).
- The consistency checker now takes the participant views directly (view validation results do not exist yet at `doParallelChecks` time), and the `ExternalCallValidator` binding is threaded through the participant wiring.
- Basic test coverage for the check outcomes and their response translation; completing the coverage is tracked in #564.

Requests without recorded external-call results short-circuit before any topology or network access, so the feature remains inert on protocol versions without external-call support and unless an extension service is configured.

## Out of scope (deliberate, per the agreed short-cut)

- Per-party verdict routing (distinguishing the affected checking parties per view) — continues in #555/#556 as further improvements, in the context of the fail-fast refactor tracked in #565.
- Rejection deduplication.
- Full test coverage — tracked in #564.

Replay disagreements within a single reinterpretation already surface as model-conformance errors through the merged #552 machinery, independently of this check.

## Coordination

With this PR the feature is functional end-to-end on `ProtocolVersion.dev`. The Daml-side PR (https://github.com/digital-asset/daml/pull/23099) can proceed once a Canton snapshot/artifact containing this PR is published, per the coordination steps in #513.

Refs #513.

### DIFF
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessingSteps.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessingSteps.scala
@@ -927,8 +927,10 @@ class TransactionProcessingSteps(
           if (malformedPayloads.nonEmpty) {
             // With malformed payloads, the responses factory takes the
             // constructResponsesForMalformedPayloads path and never awaits the check (only its
-            // alarms matter there); drain it so that failures are logged rather than silently
-            // discarded.
+            // alarms matter there). The check is currently synchronous in that case (it makes no
+            // validator calls without runValidation), so the drain only matters should the check
+            // ever become asynchronous there; it stays so that failures would be logged rather
+            // than silently discarded.
             FutureUnlessShutdownUtil.doNotAwaitUnlessShutdown(
               checkResultF,
               "external-call check failed for a request with malformed payloads",
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessingSteps.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessingSteps.scala
@@ -94,7 +94,13 @@ import com.digitalasset.canton.topology.client.TopologySnapshot
 import com.digitalasset.canton.topology.{ParticipantId, PhysicalSynchronizerId}
 import com.digitalasset.canton.tracing.TraceContext
 import com.digitalasset.canton.util.ShowUtil.*
-import com.digitalasset.canton.util.{EitherTUtil, ErrorUtil, LoggerUtil, RoseTree}
+import com.digitalasset.canton.util.{
+  EitherTUtil,
+  ErrorUtil,
+  FutureUnlessShutdownUtil,
+  LoggerUtil,
+  RoseTree,
+}
 import com.digitalasset.canton.version.ProtocolVersion
 import com.digitalasset.canton.{
   LedgerSubmissionId,
@@ -131,6 +137,7 @@ class TransactionProcessingSteps(
     createNodeEnricher: ContractEnricher,
     authorizationValidator: AuthorizationValidator,
     internalConsistencyChecker: InternalConsistencyChecker,
+    externalCallCheck: ExternalCallCheck,
     tracker: CommandProgressTracker,
     protected val loggerFactory: NamedLoggerFactory,
     futureSupervisor: FutureSupervisor,
@@ -898,12 +905,44 @@ class TransactionProcessingSteps(
             }
         )
 
+        // The external-call check re-validates recorded external-call results against the
+        // extension service. Like the model conformance check, it is kicked off here and only
+        // awaited when the confirmation responses are created, so that its network I/O runs
+        // concurrently with the remaining checks.
+        externalCallCheckResultF = {
+          val participantViews = parsedRequest.rootViewTrees.forgetNE
+            .flatMap(viewTree => viewTree.view.allSubviewsWithPosition(viewTree.viewPosition))
+            .map { case (view, viewPosition) =>
+              // The represented view of each root view tree is fully unblinded
+              // (FullTransactionViewTree invariant), so tryCreate cannot throw for it or any of
+              // its subviews (computeValidationResult below relies on the same invariant).
+              viewPosition -> checked(ParticipantTransactionView.tryCreate(view))
+            }
+            .toMap
+          val checkResultF = externalCallCheck.check(
+            requestId,
+            participantViews,
+            runValidation = malformedPayloads.isEmpty,
+          )
+          if (malformedPayloads.nonEmpty) {
+            // Responses for malformed requests are constructed without awaiting the check
+            // (only its alarms matter there); drain it so that failures are logged rather
+            // than silently discarded.
+            FutureUnlessShutdownUtil.doNotAwaitUnlessShutdown(
+              checkResultF,
+              "external-call check failed for a request with malformed payloads",
+            )
+          }
+          checkResultF
+        }
+
       } yield ParallelChecksResult(
         authenticationResult,
         consistencyResultE,
         authorizationResult,
         conformanceResultET,
         internalConsistencyResultET,
+        externalCallCheckResultF,
         timeValidationE,
         replayCheckResult,
       )
@@ -976,6 +1015,7 @@ class TransactionProcessingSteps(
         authorizationResult = parallelChecksResult.authorizationResult,
         modelConformanceResultET = parallelChecksResult.conformanceResultET,
         internalConsistencyResultET = parallelChecksResult.internalConsistencyResultET,
+        externalCallCheckResultF = parallelChecksResult.externalCallCheckResultF,
         consumedInputsOfHostedParties = usedAndCreated.contracts.consumedInputsOfHostedStakeholders,
         witnessed = usedAndCreated.contracts.witnessed,
         createdContracts = usedAndCreated.contracts.created,
@@ -1637,6 +1677,7 @@ object TransactionProcessingSteps {
         ErrorWithInternalConsistencyCheck,
         Unit,
       ],
+      externalCallCheckResultF: FutureUnlessShutdown[ExternalCallCheck.Result],
       timeValidationResultE: Either[TimeCheckFailure, Unit],
       replayCheckResult: Option[String],
   ) {
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessingSteps.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessingSteps.scala
@@ -925,9 +925,10 @@ class TransactionProcessingSteps(
             runValidation = malformedPayloads.isEmpty,
           )
           if (malformedPayloads.nonEmpty) {
-            // Responses for malformed requests are constructed without awaiting the check
-            // (only its alarms matter there); drain it so that failures are logged rather
-            // than silently discarded.
+            // With malformed payloads, the responses factory takes the
+            // constructResponsesForMalformedPayloads path and never awaits the check (only its
+            // alarms matter there); drain it so that failures are logged rather than silently
+            // discarded.
             FutureUnlessShutdownUtil.doNotAwaitUnlessShutdown(
               checkResultF,
               "external-call check failed for a request with malformed payloads",
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallCheck.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallCheck.scala
@@ -23,15 +23,16 @@ import scala.concurrent.ExecutionContext
   *
   * The check has two parts:
   *   - consistency: whether occurrences of the same external call recorded across the request agree
-  *     on their output ([[ExternalCallConsistencyChecker]]); every visible disagreement is alarmed,
+  *     on their output ([[ExternalCallConsistencyChecker]]),
   *   - re-validation: whether the (undisputed) recorded outputs agree with the extension service,
   *     re-executing each distinct call once ([[ExternalCallValidator]]).
   *
-  * The outcome is a single per-request `ExternalCallCheck.Result`: a disagreement from either part
-  * rejects the request on behalf of all hosted confirming parties, and a recorded result that
-  * cannot be re-validated leads to an abstention instead of an approval (see
-  * [[TransactionConfirmationResponsesFactory]]). Disagreements within the replay data of a single
-  * reinterpretation already surface as model-conformance errors independently of this check.
+  * Every disagreement found by either part is alarmed. The outcome is a single per-request
+  * `ExternalCallCheck.Result`: a disagreement from either part rejects the request on behalf of all
+  * hosted confirming parties, and a recorded result that cannot be re-validated leads to an
+  * abstention instead of an approval (see [[TransactionConfirmationResponsesFactory]]).
+  * Disagreements within the replay data of a single reinterpretation already surface as
+  * model-conformance errors independently of this check.
   *
   * @param externalCallValidator
   *   The validator used to re-run external calls against the extension service.
@@ -81,11 +82,9 @@ class ExternalCallCheck(
       // rejects the request, so only the party-independent visible inconsistencies are consulted.
       val consistency = ExternalCallConsistencyChecker.check(views, Set.empty)
 
-      consistency.visibleInconsistencies.foreach { inconsistency =>
-        ExternalCallValidationError.ExternalCallResultDisagreementAlarm
-          .Warn(s"Observed inconsistent external call results: ${inconsistency.description}")
-          .logWithContext(Map("requestId" -> requestId.toString))
-      }
+      // Every visible inconsistency is alarmed: only the first one is propagated into the
+      // rejection, so the confirmation-responses factory reports just that one.
+      consistency.visibleInconsistencies.foreach(alarmDisagreement(requestId, _))
 
       // The checker sorts the inconsistencies, so the reported disagreement is deterministic.
       consistency.visibleInconsistencies.headOption match {
@@ -94,7 +93,7 @@ class ExternalCallCheck(
         case None if !runValidation =>
           FutureUnlessShutdown.pure(Passed)
         case None =>
-          validateRecordedResults(recordedResults)
+          validateRecordedResults(requestId, recordedResults)
       }
     }
   }
@@ -106,7 +105,8 @@ class ExternalCallCheck(
     * deterministic.
     */
   private def validateRecordedResults(
-      recordedResults: Seq[(ViewPosition, ViewParticipantData.ViewExternalCallResult)]
+      requestId: RequestId,
+      recordedResults: Seq[(ViewPosition, ViewParticipantData.ViewExternalCallResult)],
   )(implicit
       traceContext: TraceContext,
       ec: ExecutionContext,
@@ -132,20 +132,25 @@ class ExternalCallCheck(
             .map(outcome => (key, occurrencesAndOutputs, recordedOutput, outcome))
       }
       .map { outcomes =>
-        val rejectionO = outcomes.collectFirst {
+        val mismatches = outcomes.collect {
           case (
                 key,
                 occurrencesAndOutputs,
                 recordedOutput,
                 validated: ExternalCallValidator.Mismatched,
               ) =>
-            val inconsistency = Inconsistency(
+            Inconsistency(
               key,
               Set(validated.computedOutput, recordedOutput),
               occurrencesAndOutputs.map { case (occurrence, _) => occurrence }.toSet,
             )
-            Rejected(inconsistency.description)
         }
+        // Mirrors the visible-inconsistency alarms in check: every disagreement with the
+        // extension service is suspicious, not only the first one.
+        mismatches.foreach(alarmDisagreement(requestId, _))
+
+        val rejectionO =
+          mismatches.headOption.map(inconsistency => Rejected(inconsistency.description))
         val abstentionO = outcomes.collectFirst {
           case (_, _, _, ExternalCallValidator.UnableToValidate(reason)) =>
             CannotValidate(reason)
@@ -153,6 +158,13 @@ class ExternalCallCheck(
         rejectionO.orElse(abstentionO).getOrElse(Passed)
       }
   }
+
+  private def alarmDisagreement(requestId: RequestId, inconsistency: Inconsistency)(implicit
+      traceContext: TraceContext
+  ): Unit =
+    ExternalCallValidationError.ExternalCallResultDisagreementAlarm
+      .Warn(s"Observed inconsistent external call results: ${inconsistency.description}")
+      .logWithContext(Map("requestId" -> requestId.toString))
 }
 
 object ExternalCallCheck {
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallCheck.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallCheck.scala
@@ -1,0 +1,180 @@
+// Copyright (c) 2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.digitalasset.canton.participant.protocol.validation
+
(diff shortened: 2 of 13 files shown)


matthiasS-da:
Start reviewing...

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallCheck.scala
```diff
@@ -0,0 +1,180 @@
...
+  *   - consistency: whether occurrences of the same external call recorded across the request agree
+  *     on their output ([[ExternalCallConsistencyChecker]]); every visible disagreement is alarmed,
+  *   - re-validation: whether the (undisputed) recorded outputs agree with the extension service,
+  *     re-executing each distinct call once ([[ExternalCallValidator]]).
+  *
+  * The outcome is a single per-request `ExternalCallCheck.Result`: a disagreement from either part
+  * rejects the request on behalf of all hosted confirming parties, and a recorded result that
+  * cannot be re-validated leads to an abstention instead of an approval (see
+  * [[TransactionConfirmationResponsesFactory]]). Distinguishing the affected checking parties per
+  * view is a tracked follow-up; disagreements within the replay data of a single reinterpretation
```
matthiasS-da: MINOR: Remove the statement about the follow-up, so we do not need a reminder to remove it later on.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallCheck.scala
```diff
@@ -0,0 +1,180 @@
...
+class ExternalCallCheck(
+    externalCallValidator: ExternalCallValidator,
+    externalCallValidationParallelism: PositiveInt,
+    protected val loggerFactory: NamedLoggerFactory,
+) extends NamedLogging {
+
+  import ExternalCallCheck.*
+
+  /** Checks consistency of the recorded external-call results and re-validates them against the
+    * extension service. The network I/O runs concurrently with the other validation suites; the
```
matthiasS-da: MINOR: Remove the sentence "The network I/O", as this is not a property of the method itself, but a consequence of how the method is used.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -244,13 +246,31 @@ class TransactionConfirmationResponsesFactory(
...
+                externalCallCheckResult match {
+                  case ExternalCallCheck.Rejected(description) =>
+                    Some(
+                      logged(
+                        requestId,
+                        LocalRejectError.ConsistencyRejections.ExternalCallResultDisagreement
+                          .Reject(description),
+                      ).toLocalReject(protocolVersion)
+                    )
+                  case _ => None
```
matthiasS-da: Please collect CannotValidate as well.

MINOR: Please replace _ by like _: Passed to make explicit what is ignored here.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -266,7 +286,25 @@ class TransactionConfirmationResponsesFactory(
                   )
```
matthiasS-da, line 302: Apparently, the abstain case is missing in the above code (6 lines above). Could you please alter the localReject case to:

case nonPositive: NonPositiveLocalVerdict if hostedConfirmingParties.nonEmpty =>
nonPositive -> hostedConfirmingParties

So that LocalAbstain gets propagted, too?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -266,7 +286,25 @@ class TransactionConfirmationResponsesFactory(
...
+                        LocalAbstainError.CannotPerformAllValidations.Abstain(reason),
+                      ).toLocalAbstain(protocolVersion) -> parties
+                    case _ => approve -> parties
+                  }
+                case other => other
+              }
+
+              localVerdictAndPartiesWithAbstentionO.map { case (localVerdict, parties) =>
+                // The abstention above preserves the tryCreate invariant: it is not malformed,
+                // and it keeps the approval's non-empty party set.
```
matthiasS-da: Please undo this chunk, and merge it into the chunk above handling Rejected.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallCheck.scala
```diff
@@ -0,0 +1,180 @@
...
+    else {
+      // Hosted parties are irrelevant for the request-level outcome: any visible disagreement
+      // rejects the request, so only the party-independent visible inconsistencies are consulted.
+      val consistency = ExternalCallConsistencyChecker.check(views, Set.empty)
+
+      consistency.visibleInconsistencies.foreach { inconsistency =>
+        ExternalCallValidationError.ExternalCallResultDisagreementAlarm
+          .Warn(s"Observed inconsistent external call results: ${inconsistency.description}")
+          .logWithContext(Map("requestId" -> requestId.toString))
+      }
```
matthiasS-da: I think this is duplicative alerting, as TransactionConfirmationResponsesFactory already create alerts. If you agree, please remove it.

matthiasS-da: I notice that TransactionConfirmationResponseFactory only alerts on the head of this collection. So this statement is not redundant and should stay.

Could you please leave a comment explaining this?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallCheck.scala
```diff
@@ -0,0 +1,180 @@
...
+          // All occurrences of the key record the same output (see the method contract).
+          val recordedOutput = occurrencesAndOutputs.headOption
+            .map { case (_, output) => output }
+            .getOrElse(ErrorUtil.invalidState("a grouped key always has at least one occurrence"))
+          externalCallValidator
+            .validate(key, recordedOutput)
+            .map(outcome => (key, occurrencesAndOutputs, recordedOutput, outcome))
+      }
+      .map { outcomes =>
+        val rejectionO = outcomes.collectFirst {
```
matthiasS-da: Following the same logic as for visibleInconsistencies, shouldn't we alert here?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallCheck.scala
angelol: Removed in 21eb080b56.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallCheck.scala
angelol: Removed in 21eb080b56 (same commit as the class-doc trim above). The call site in `doParallelChecks` keeps the comment explaining why it is kicked off there.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: Done in 070a46f8ef: `case ExternalCallCheck.Passed | ExternalCallCheck.CannotValidate(_) => None` (the outcome is named `CannotValidate` on this ADT; `Passed` is a case object, hence the object reference rather than a type ascription). The abstention match introduced in ab72da138e spells out its ignored outcomes the same way.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: Done in ab72da138e. The abstain case was previously handled a few lines below this (`localVerdictAndPartiesWithAbstentionO`, downgrading a selected `LocalApprove`); with your suggestion the abstention is instead appended last to `localVerdicts` and propagates through the same `collectFirst`, so that extra mapping step is gone. Every rejection still ranks ahead of the abstention via the chain order — 9ca43ec019 adds a test pinning that precedence. One behavioral note: the `CANNOT_PERFORM_ALL_VALIDATIONS` INFO line is now logged per view whenever re-validation was impossible, even where a rejection wins the selection, matching how the sibling rejection vals log.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessingSteps.scala
```diff
@@ -898,12 +905,44 @@ class TransactionProcessingSteps(
...
+          )
+          if (malformedPayloads.nonEmpty) {
+            // Responses for malformed requests are constructed without awaiting the check
+            // (only its alarms matter there); drain it so that failures are logged rather
+            // than silently discarded.
+            FutureUnlessShutdownUtil.doNotAwaitUnlessShutdown(
+              checkResultF,
+              "external-call check failed for a request with malformed payloads",
+            )
+          }
```
matthiasS-da, line 938: I don't get why we need this doNotAwait here. We normally use this for discarded future, but checkResultF is propagated to TransactionConfirmationResponsesProcessor, awaited there, alerted and the result is converted into a verdict.

Please double-check and let me know if you agree.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -244,21 +246,55 @@ class TransactionConfirmationResponsesFactory(
...
+                externalCallCheckResult match {
+                  case ExternalCallCheck.CannotValidate(reason) =>
+                    Some(
+                      logged(
+                        requestId,
+                        LocalAbstainError.CannotPerformAllValidations.Abstain(reason),
+                      ).toLocalAbstain(protocolVersion)
+                    )
+                  case ExternalCallCheck.Passed | ExternalCallCheck.Rejected(_) => None
+                }
```
matthiasS-da, line 273: MINOR: Combine the two above match statements into a single one. :)

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -244,21 +246,55 @@ class TransactionConfirmationResponsesFactory(
...
-                  replayRejections
+                  replayRejections ++ externalCallAbstentions
 
               val localVerdictAndPartiesO = localVerdicts
                 .collectFirst[(LocalVerdict, Set[LfPartyId])] {
                   case malformed: LocalReject if malformed.isMalformed => malformed -> Set.empty
-                  case localReject: LocalReject if hostedConfirmingParties.nonEmpty =>
-                    localReject -> hostedConfirmingParties
+                  case nonPositive: NonPositiveLocalVerdict if hostedConfirmingParties.nonEmpty =>
+                    nonPositive -> hostedConfirmingParties
```
matthiasS-da: Right, now I see why you have two match statements above.

We actually need separate case entries for LocalReject and LocalAbstain here so that LocalRejects gets higher precedence.

Happy to clean this up post-merge, if that saves us some time.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
matthiasS-da: Thanks. See my separate follow-up comments.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallProtocolIntegrationTest.scala
```diff
@@ -0,0 +1,401 @@
...
+  * and their translation into confirmation responses. Completing the coverage is a tracked
+  * post-merge task.
+  */
+final class ExternalCallProtocolIntegrationTest
+    extends BaseTestWordSpec
+    with HasExecutionContext
+    with ExternalCallValidationTestUtil {
+
+  protected val factory: ExampleTransactionFactory =
+    new ExampleTransactionFactory(versionOverride = Some(ProtocolVersion.dev))(
```
matthiasS-da, line 59: We shouldn't hardcode the protocol version as already pointed out in previous PRs.

I'm ok if you address that as a post-merge cleanup.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallProtocolIntegrationTest.scala
```diff
@@ -0,0 +1,401 @@
...
+  ): ExternalCallCheck.Result =
+    externalCallCheck(validator).check(requestId, views, runValidation).futureValueUS
+
+  private def assertDisagreementAlarm[A](within: => A): A =
+    loggerFactory.assertLogs(
+      within,
+      (logEntry: LogEntry) => {
+        logEntry.shouldBeCantonErrorCode(
+          ExternalCallValidationError.ExternalCallResultDisagreementAlarm
+        )
```
matthiasS-da, line 145: As a post-merge cleanup, we should call shouldBeCantonError to check the error message as well.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: CannotValidate is now collected in the same match (2c157386f5, together with the merge asked for in the thread below).

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: Done in 2c157386f5: one `externalCallVerdicts` match now produces the external-call verdict (reject or abstain) at the former rejections slot, and the separate mapping step is gone. This supersedes the ordering note in the thread above: the abstention now ranks like the rejection, so verdicts from the earlier suites (activeness, time, contract consistency) still win against it, while verdicts from the later suites (authentication, authorization, model conformance, internal consistency, replay — including their malformed rejections, when confirming parties are hosted) are shadowed by it. A test pins that ordering in both directions, and the replay-rejection comment ("or something at least as strong") is adjusted for the abstention case.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallCheck.scala
angelol: Comment added in 5f7e9d6d07 ("Every visible inconsistency is alarmed: only the first one is propagated into the rejection, so the confirmation-responses factory reports just that one.").

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallCheck.scala
angelol: Agreed — done in 5f7e9d6d07: every re-validation disagreement is alarmed (not only the first), through the same helper as the visible inconsistencies, and the mismatch test now asserts the alarm.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessingSteps.scala
angelol: Double-checked — the drain is needed, but only for the malformed-payloads path, and it is guarded accordingly. `createConfirmationResponses` returns `ProcessingSteps.constructResponsesForMalformedPayloads` when `malformedPayloads` is non-empty (TransactionConfirmationResponsesFactory.scala:328-341) and never touches the `TransactionValidationResult` futures on that path; `responsesForWellFormedPayloads`, which awaits the check, is only entered otherwise. On the well-formed path nothing is drained and the factory is the sole consumer, as you describe. Sharpened the comment to name that branch (5a44ee2ae7).

matthiasS-da: In the malformed-payloads path, we have `!runValidation`. Hence, all code run as part of `ExternalCallChecks.check` is synchronous. Therefore, no drain is needed.

However, if the implementation of `check` ever changes to be asynchronous in the `!runValidation` case, we would need the drain. So let's leave it there.

Do you agree with my reasoning?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
```diff
@@ -187,7 +189,8 @@ class TransactionConfirmationResponsesFactory(
                       requestId,
                       // Conceptually, a normal LocalReject for the admin party should suffice for rejecting replays.
                       // However, we nevertheless use a `Malformed` rejection here so that the rejection preference sorting
-                      // ensures that this rejection or something at least as strong will make it to the mediator.
+                      // ensures that this rejection, something at least as strong, or the external-call
+                      // abstention (which ranks earlier in the verdicts below) will make it to the mediator.
```
matthiasS-da: This change looks like an accident. Undo?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: Already done in 2c157386f5 — our pushes crossed with your comments; the two matches are one `externalCallVerdicts` match now.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: Done now rather than post-merge, in c9cb6d63e1: the selection runs in two passes — the first is exactly the previous rejection-only `collectFirst`, and only if no rejection matched anywhere in the verdicts does a second `collectFirst` consider the abstention. Separate case entries within a single `collectFirst` would not achieve this, since the cases are tried per element in list order (an earlier abstain element would still win over a later reject). With rejections taking precedence unconditionally again, the replay comment ("or something at least as strong") holds as originally written and is reverted; the two precedence tests now pin a rejection beating the abstention from both an earlier suite (time) and a later one (a malformed authorization rejection, selected with empty parties as before).

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallProtocolIntegrationTest.scala
angelol: Taking the post-merge offer — tracked in #564 (https://github.com/digital-asset/canton/issues/564#issuecomment-4902335616), together with the other test completions.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallProtocolIntegrationTest.scala
angelol: Tracked in #564 as well (https://github.com/digital-asset/canton/issues/564#issuecomment-4902335616).

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
matthiasS-da: Note that right now LocalAbstain from external calls override LocalRejects from other sources. So better clean this up before merging.

matthiasS-da (dismissed):
LGTM except for the post-merge cleanup requests.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessingSteps.scala
angelol: Agreed, that is exactly right: all three `!runValidation` exits of `check` are `FutureUnlessShutdown.pure`, so the future handed to the drain is already completed and the drain is currently vacuous — it only matters should the check ever become asynchronous on this path. Left it in place as you suggest and captured your reasoning in the comment (509ede2ddb).

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: Not an accident, but already reverted in c9cb6d63e1 (our pushes crossed again): the adjustment tracked the interim ordering where the abstention could shadow the replay rejection; with rejections taking precedence again the original claim holds and the comment is back to its upstream wording.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/TransactionConfirmationResponsesFactory.scala
angelol: Done before merge in c9cb6d63e1, pushed just as you wrote this — rejections from all sources take precedence over the abstention again (two-pass selection, see the reply above).

matthiasS-da (dismissed):
LGTM subject to the agreed post-cleanup tasks.

matthiasS-da:
@angelol there is a problem with formatting. Let me take care of it.

matthiasS-da:
> @angelol there is a problem with formatting. Let me take care of it.

@angelol I can't, as I don't have permissions. Please run `sbt format`, commit and push. Afterwards, I'll merge the PR.

angelol:
Done — `sbt format` fixed an import-order slip in the test file (my per-module scalafmt runs missed the scalafix part; noted for next time). Pushed as 9eef119606; the branch is quiet from my side. Thanks for the fast turnaround today!

matthiasS-da (approved):

PR #571, opened 2026-07-08
refactor(external-call): move ExternalCallKey to community/base

angelol:
Pure move, no behavior change: relocate `ExternalCallKey` and the `ExternalCallPayloadDescription` payload-size helper from the participant `DAMLe` engine wrapper into `community/base` (`com.digitalasset.canton.data`).

## Why

`ExternalCallKey` is a protocol-data identity (four canonical strings plus an ordering) derived from `ViewParticipantData.ViewExternalCallResult`, which already lives in `base`. Moving it there lets `TransactionView.validated` fail fast on a recorded-result disagreement in the follow-up — the centralization requested at https://github.com/digital-asset/canton/pull/552#discussion_r3506482637 — and keeps the "same external call" identity in one place rather than split across `DAMLe` and the participant validators.

## On the target package (r3506626583)

The suggestion at https://github.com/digital-asset/canton/pull/552#discussion_r3506626583 was to give `ExternalCallKey` its own file under `com.digitalasset.canton.participant.protocol.validation`. This PR instead puts it in `com.digitalasset.canton.data` (base), because the fail-fast centralization (r3506482637) requires `TransactionView.validated` — which lives in `community/base` — to compute `ExternalCallReplayData` keyed by `ExternalCallKey`. `base` cannot depend on `participant`, so the key has to live in `base` for that step to be possible. This honors the encapsulation intent of the original comment (own file, no longer buried in `DAMLe`) at the location the architectural change dictates. Nesting the key privately inside `ExternalCallReplayData` is not viable — the consistency checker and validators group and order by it.

`ExternalCallReplayData` and its `ReinterpretationError`-typed disagreement stay in `DAMLe` for now; they move to `base` (collapsed to a single output per key) in the fail-fast change.

## Scope

- New `community/base/.../data/ExternalCallKey.scala` and `.../data/ExternalCallPayloadDescription.scala` (verbatim moves; `private[canton]` visibility preserved).
- References repointed in `DAMLe`, `ExternalCallCheck`, `ExternalCallConsistencyChecker`, `ExternalCallValidator`, `ExtensionServiceExternalCallValidator`, and their tests.

## Testing

`community-participant/Test/compile` green; external-call suites green at `CANTON_PROTOCOL_VERSION=dev` (`DAMLeExternalCallTest`, `ExternalCallConsistencyCheckerTest`, `ExtensionServiceExternalCallValidatorTest`, `ExternalCallProtocolIntegrationTest` — 29/0/0).

[...]

PR #572, opened 2026-07-09
feat(external-call): fail fast on recorded-result disagreements in TransactionView

angelol:
Centralizes the recorded external-call disagreement check in `TransactionView.validated`, per the #552 review ([r3506482637](https://github.com/digital-asset/canton/pull/552#discussion_r3506482637)) — moving from the fail-last model (a disagreement surfaced only during reinterpretation, as a model-conformance error) to fail-fast.

> Rebased onto `main` after #571 merged; the diff is now exactly this PR's change.

## What changes

- `TransactionView` computes `ExternalCallReplayData` over its subtree and **fails the view as malformed** if the same semantic key is recorded with conflicting outputs — so a view carrying an ambiguous recorded result is never reinterpreted. The aggregation recurses via the subviews' memoized values (`ExternalCallReplayData.merge`), so each view pays only for its direct children plus its own results, and views without results — every view below protocol version dev — cost effectively nothing. Exposed as a private `…E` used by `validated` plus a total accessor for consumers (the accessor skips blinded participant data, like the old `ModelConformanceChecker` aggregation did).
- `ExternalCallReplayData` moves to `community/base` as **single output per key** (`Map[ExternalCallKey, Bytes]`), built with `MapsUtil.toNonConflictingMap`.
- Removed: `ExternalCallRecordedResultDisagreement`, the multi-valued replay data and the late `outputFor` disagreement branch in `DAMLe`, and `ModelConformanceChecker.externalCallReplayDataFor` (now consumes the view's total `externalCallReplayData`).
- The conflict message is **size-only, never the payload bytes** (it is logged downstream).
- **Submission side**: `NextGenTransactionTreeFactory` builds views via `TransactionView.create` and surfaces a validation failure as a typed `InvalidTransactionViewError` (soft error), so a transaction whose recorded results conflict — reachable from a non-deterministic extension service or a crafted interactive-submission execute request — rejects gracefully instead of escaping as an uncaught `InvalidView`. (The legacy factory records no external-call results and is unchanged.)

## Behavioural note (verdict path for cross-view disagreements)

The check is grouped with the participant-data validation in `TransactionView.validated`. A conflict *within a single view* is caught cleanly at deserialization (→ `MalformedPayload`). A **cross-view** conflict (same key in a parent and a child view) is not visible until light-view-tree reassembly, where `validated` runs over the unblinded subtree and throws `InvalidView` at `LightTransactionViewTree.toFullViewTrees`.

This is the **same path the existing cross-view checks already take** — `validateViewParticipantData`'s package-preference check and `validateViewCommonData`'s equal-`viewCommonData` check are both cross-view, both only fire once subviews are unblinded, and both throw `InvalidView` at reassembly (no `catchOnly` there today), aborting the request. The transaction is not committed either way. So this PR inherits the established behaviour rather than introducing a new one; it does not add a `catchOnly`/`MalformedPayload` conversion at reassembly. Happy to follow up with one (it would give all three cross-view checks a clean malformed reject) if you'd prefer.

## Testing

`community-common` / `community-participant` / `community-app` Test compile green; `sbt format` clean. At `CANTON_PROTOCOL_VERSION=dev`: new `ExternalCallReplayDataTest` (`fromResults` + `merge`, conflict message pinned in full, payload no-leak asserted against the hex renderings), `TransactionViewTest` incl. within-view and across-parent/subview conflict cases (the cross-view fixture keeps the child's `viewCommonData` distinct, so it isolates the external-call rejection), `NextGenTransactionTreeFactoryTest` incl. the typed conflicting-results rejection, `ModelConformanceCheckerTest`, `ExampleTransactionConformanceTest`, `ExternalCallProtocolIntegrationTest`, `DAMLeExternalCallTest`, `SerializationDeserializationTest` — all green. `GeneratorsData` never generates a same-key/different-output view but keeps the legal same-key/same-output multiplicity in round-trip coverage.

### DIFF
--- a/community/base/src/main/scala/com/digitalasset/canton/data/TransactionView.scala
+++ b/community/base/src/main/scala/com/digitalasset/canton/data/TransactionView.scala
@@ -241,7 +241,8 @@ final case class TransactionView private (
   /** The single recorded output per external-call key, aggregated over this view and its subviews.
     *
     * @throws java.lang.IllegalStateException
-    *   if the same key was recorded with conflicting outputs, which a validated view cannot contain
+    *   if the [[ViewParticipantData]] of this view or any subview is blinded, or if the same key
+    *   was recorded with conflicting outputs, which a validated view cannot contain
     */
   def externalCallReplayData(implicit
       loggingContext: NamedLoggingContext
@@ -254,10 +255,25 @@ final case class TransactionView private (
     inputsAndCreatedE.map(_._2)
 
   private lazy val externalCallReplayDataE: Either[String, ExternalCallReplayData] =
-    // Recurses via the subviews' memoized values, so the aggregation costs each view only its
-    // direct children plus its own results, and stays trivial for views without any results.
+    for {
+      vpd <- unblindViewParticipantData("External-call replay data")
+      _ <- subviews.allUnblinded(hash =>
+        s"External-call replay data of view $viewHash can be computed only if all subviews are unblinded, but $hash is blinded"
+      )
+      subviewData <- subviews.unblindedElements.traverse(_.externalCallReplayDataE)
+      merged <- ExternalCallReplayData.merge(subviewData, vpd.externalCallResults.map(_.result))
+    } yield merged
+
+  /** Conflict check for `validated` over the visible parts of the subtree: blinded participant data
+    * and blinded subviews contribute no results, like the other `validated` checks, which also skip
+    * blinded data — `validated` runs while blinding, so it must tolerate blinded parts.
+    * [[externalCallReplayDataE]] instead reports blinded data as an error, so replay never runs on
+    * silently-partial data. Both recurse via the subviews' memoized values, so the aggregation
+    * costs each view only its direct children plus its own results.
+    */
+  private lazy val visibleExternalCallReplayDataE: Either[String, ExternalCallReplayData] =
     subviews.unblindedElements
-      .traverse(_.externalCallReplayDataE)
+      .traverse(_.visibleExternalCallReplayDataE)
       .flatMap(subviewData =>
         ExternalCallReplayData.merge(
           subviewData,
@@ -384,7 +400,7 @@ final case class TransactionView private (
       // Grouped with the participant-data validation above: a key recorded with conflicting
       // outputs across this view's subtree makes the view malformed, because reinterpretation
       // cannot proceed on an ambiguous recorded result.
-      _ <- externalCallReplayDataE
+      _ <- visibleExternalCallReplayDataE
       _ <- viewCommonData.unwrap match {
         case Left(_) => Either.unit
         case Right(d) => validateViewCommonData(d, childCommonData)
--- a/community/base/src/main/scala/com/digitalasset/canton/data/TransactionView.scala
+++ b/community/base/src/main/scala/com/digitalasset/canton/data/TransactionView.scala
@@ -244,7 +244,7 @@ final case class TransactionView private (
     *   if the [[ViewParticipantData]] of this view or any subview is blinded, or if the same key
     *   was recorded with conflicting outputs, which a validated view cannot contain
     */
-  def externalCallReplayData(implicit
+  def tryExternalCallReplayData(implicit
       loggingContext: NamedLoggingContext
   ): ExternalCallReplayData = getOrError(externalCallReplayDataE)
 
--- a/community/base/src/main/scala/com/digitalasset/canton/data/TransactionView.scala
+++ b/community/base/src/main/scala/com/digitalasset/canton/data/TransactionView.scala
@@ -238,11 +238,27 @@ final case class TransactionView private (
       loggingContext: NamedLoggingContext
   ): Map[LfContractId, CreatedContractInView] = getOrError(createdContractsE)
 
+  /** The single recorded output per external-call key, aggregated over this view and its subviews.
+    *
+    * @throws java.lang.IllegalStateException
+    *   if the same key was recorded with conflicting outputs, which a validated view cannot contain
+    */
+  def externalCallReplayData(implicit
+      loggingContext: NamedLoggingContext
+  ): ExternalCallReplayData = getOrError(externalCallReplayDataE)
+
   private def inputContractsE: Either[String, Map[LfContractId, InputContract]] =
     inputsAndCreatedE.map(_._1)
 
   private def createdContractsE: Either[String, Map[LfContractId, CreatedContractInView]] =
     inputsAndCreatedE.map(_._2)
+
+  private lazy val externalCallReplayDataE: Either[String, ExternalCallReplayData] =
+    ExternalCallReplayData.fromResults(
+      flatten
+        .flatMap(_.viewParticipantData.unwrap.toOption.toList.flatMap(_.externalCallResults))
+        .map(_.result)
+    )
 
   private lazy val inputsAndCreatedE: Either[
     String,
@@ -358,6 +374,10 @@ final case class TransactionView private (
         case Right(d) =>
           validateViewParticipantData(d, childParticipantData, inputContractsO)
       }
+      // Grouped with the participant-data validation above: a key recorded with conflicting
+      // outputs across this view's subtree makes the view malformed, because reinterpretation
+      // cannot proceed on an ambiguous recorded result.
+      _ <- externalCallReplayDataE
       _ <- viewCommonData.unwrap match {
         case Left(_) => Either.unit
         case Right(d) => validateViewCommonData(d, childCommonData)
--- a/community/base/src/main/scala/com/digitalasset/canton/data/TransactionView.scala
+++ b/community/base/src/main/scala/com/digitalasset/canton/data/TransactionView.scala
@@ -254,11 +254,18 @@ final case class TransactionView private (
     inputsAndCreatedE.map(_._2)
 
   private lazy val externalCallReplayDataE: Either[String, ExternalCallReplayData] =
-    ExternalCallReplayData.fromResults(
-      flatten
-        .flatMap(_.viewParticipantData.unwrap.toOption.toList.flatMap(_.externalCallResults))
-        .map(_.result)
-    )
+    // Recurses via the subviews' memoized values, so the aggregation costs each view only its
+    // direct children plus its own results, and stays trivial for views without any results.
+    subviews.unblindedElements
+      .traverse(_.externalCallReplayDataE)
+      .flatMap(subviewData =>
+        ExternalCallReplayData.merge(
+          subviewData,
+          viewParticipantData.unwrap.toOption.toList
+            .flatMap(_.externalCallResults)
+            .map(_.result),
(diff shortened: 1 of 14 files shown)


community/base/src/main/scala/com/digitalasset/canton/data/TransactionView.scala
```diff
@@ -238,12 +238,35 @@ final case class TransactionView private (
...
 
+  private lazy val externalCallReplayDataE: Either[String, ExternalCallReplayData] =
+    // Recurses via the subviews' memoized values, so the aggregation costs each view only its
+    // direct children plus its own results, and stays trivial for views without any results.
+    subviews.unblindedElements
+      .traverse(_.externalCallReplayDataE)
+      .flatMap(subviewData =>
+        ExternalCallReplayData.merge(
+          subviewData,
+          viewParticipantData.unwrap.toOption.toList
```
matthiasS-da, line 280: Right now, a blinded viewParticipantData will get silently ignored. Could you please return Left(error) instead?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
```diff
@@ -851,6 +855,24 @@ class NextGenTransactionTreeFactory(
 
 object NextGenTransactionTreeFactory {
 
+  /** A transaction that records conflicting outputs for the same external call cannot form a valid
+    * transaction tree: `TransactionView.validated` fails any view whose subtree contains the
+    * conflict. Reject the transaction with a typed error before view construction instead of
+    * hitting the factory's view-invariant check.
+    */
```
matthiasS-da: I think we don't need this check as it is redundant, but happy to discuss. Why is it preferable to catch this here instead of when constructing TransactionView?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/TransactionTreeFactory.scala
```diff
@@ -208,6 +208,16 @@ object TransactionTreeFactory {
...
 
+  /** Indicates that the submitted transaction records conflicting outputs for the same external
+    * call, so no valid transaction tree can be formed from it.
+    */
+  final case class ConflictingExternalCallResultsError(message: String)
+      extends TransactionTreeConversionError {
+    override protected def pretty: Pretty[ConflictingExternalCallResultsError] = prettyOfClass(
+      unnamedParam(_.message.unquoted)
+    )
+  }
```
matthiasS-da, line 219: This may become obsolete, depending of the outcome of the previous comment.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
```diff
@@ -314,8 +294,7 @@ class ModelConformanceChecker(
       view.viewParticipantData.tryUnwrap.keyResolution.fmap(_.unversioned.contracts),
     )
 
-    lazy val externalCallReplayData: ExternalCallReplayData =
-      externalCallReplayDataFor(view)
+    lazy val externalCallReplayData = view.externalCallReplayData
```
matthiasS-da: MINOR: Let's please inline this.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactoryTest.scala
```diff
@@ -343,6 +343,35 @@ final class NextGenTransactionTreeFactoryTest
...
+              ),
+              successfulLookup(example),
+            ).value.map { result =>
+              result.left.value shouldBe ConflictingExternalCallResultsError(
+                "externalCallResults records conflicting outputs for the same external call: " +
+                  "ExternalCallKey(extension id = \"extension\", function id = \"function\", " +
+                  "config bytes = \"6 bytes\", input bytes = \"5 bytes\") with outputs [8 bytes, 8 bytes]"
+              )
+            }
+          }
```
matthiasS-da, line 373: I wonder how this test would look if the redundant check in TransactionTreeFactory is removed.

matthiasS-da (commented):
Thanks for the effort - I've left a few comments, but overall the direction is great!

community/base/src/main/scala/com/digitalasset/canton/data/TransactionView.scala
angelol: Done in 14afee79ca. Two things fell out of implementing it:

- The accessor path now mirrors `inputsAndCreatedE`: `externalCallReplayDataE` returns `Left` when this view's participant data is blinded — and also when any subview is blinded, since the aggregation spans the subtree — so replay can never run on silently-partial data.
- `validated` keeps a visible-parts-only conflict check (`visibleExternalCallReplayDataE`): `validated` runs while blinding (`withBlindedSubtrees` and `tryBlindForTransactionViewTree` construct views via `tryCreate`), so a `Left` for blinded data there would make blinding itself throw `InvalidView`. Skipping blinded data also matches the neighboring `validated` checks (package preferences, equal `viewCommonData`).

Added tests: a view with a blinded subview validates but `externalCallReplayData` refuses with the blinded-data error, and a view with blinded participant data validates but likewise refuses the replay data.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
angelol: It's redundant as a detector — the `TransactionView.validated` check fires on the same input — but the two failure modes differ. The factories call the throwing `tryCreate` overload inside an `EitherT` yield, and nothing on that path catches `InvalidView` (`TransactionConfirmationRequestFactory` only `leftMap`s the typed channel), so without the pre-check a conflict surfaces as a failed future / internal error at phase 1. And unlike the other `validated` failures reachable from the factory (equal `viewCommonData`, package preferences), which are genuine factory invariants and should crash loudly, this one is data-dependent: an extension service answering the same call differently within one command, or a crafted `execute` request (`PreparedTransactionDecoder` checks hex and VDev-gating, but not cross-occurrence output consistency), reaches it from outside. Before this PR the same input built the tree and was rejected in phase 3 by model conformance, so dropping the check would turn that rejection into an internal error.

Catching at the construction site instead (`TransactionView.create` + `leftMap` into a typed error) also works, but it converts the genuine invariant violations into soft errors as well — that's why I preferred the explicit pre-check; it is one pass over the exercise nodes and also rejects a cross-sibling-root conflict at submission, which per-view `validated` cannot see. Happy to switch to catch-at-construction or drop the check if you prefer accepting the internal error there.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/TransactionTreeFactory.scala
angelol: Kept for now pending the discussion on the pre-check — if that goes a different way, this error variant goes with it.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ModelConformanceChecker.scala
angelol: Done in 4020afdcb7.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactoryTest.scala
angelol: Without the pre-check it could no longer assert a rejection: the conflict would surface as a failed future from the uncaught `InvalidView` inside the factory, so the test would assert an internal error instead of a typed `Left`. That contrast is essentially why the check exists — details in the thread on the pre-check.

community/base/src/main/scala/com/digitalasset/canton/data/TransactionView.scala
angelol: (After rebasing onto `main` following #571's merge, the commit referenced above is now 927ff6294e.)

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
matthiasS-da: Thanks for the explanation - really appreciated! 

I agree that a crafted `execute` request can trigger that and therefore it should fail gracefully, without an external error.

From my point of view, it is more of an "accident" that the call to TransactionView.tryCreate has worked so far. My preferred solution is to call `TransactionView.create` instead and go for a "soft" error (i.e. Left(...)) by default. Could you please give this a try?

If this is too difficult to achieve (failed regression tests), I'm happy to go for the internal error instead.

In any case, please remove the redundant error handling from TransactionTreeFactory.

community/base/src/main/scala/com/digitalasset/canton/data/TransactionView.scala
matthiasS-da: That looks good and thanks for working out these details!

One minor thing: could you please rename externalCallReplayData to tryExternalCallReplayData to flag that this could throw?

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
angelol: Done in 97ed17b93e — `create` worked out cleanly, no regression-test fallout. The factory now surfaces a validation failure as a typed `InvalidTransactionViewError` (generalized from the conflict-specific variant, since the soft error can now carry any `validated` message), and the pre-check plus its error are removed from `TransactionTreeFactory`. The legacy factory records no external-call results, so I left its `tryCreate` untouched — happy to convert it the same way if you'd like. The conflicting-results factory test now asserts the same message through the new error, which also answers your question below.

community/base/src/main/scala/com/digitalasset/canton/data/TransactionView.scala
angelol: Done in 7f37827f02.

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/submission/NextGenTransactionTreeFactory.scala
matthiasS-da: Happy to leave the legacy factory as is. Let's not risk any breaking changes.

matthiasS-da (approved):
LGTM

PR #553, opened 2026-06-30
feat(external-call): extension-service handler, config and command-execution wiring

angelol:
## Summary

This is PR 5 of 8 in the external-call runtime-integration series tracked in #513, stacked on PR 4.

The external-call stack (#513): #506 added the transaction-side representation for recording external-call results; #514 added the LF `EXTERNAL_CALL` builtin surface; #518 wired it into Speedy and the LF engine; #522 added transaction protobuf encoding/decoding; #526 added Canton protocol serialization for recorded results; #537 added the participant-side extension-service client. #541 implemented the remaining runtime integration as one PR; this series splits #541 into 8 independently reviewable PRs (each < 1000 lines) and supersedes it.

This PR connects the #537 extension-service client to command execution: it adds the handler that routes engine external-call questions to configured extension services, the participant configuration that keys services by extension id, and the node wiring.

## Scope

This PR adds:

- the `ExtensionServiceExternalCallHandler` implementation of the PR 3 handler SPI, dispatching engine external-call questions to the #537 extension-service client (`ExtensionServiceManager` / `HttpExtensionServiceClient` / `ExtensionServiceClient` glue)
- participant extension-service configuration keyed by extension id (`CantonConfig`, `ConfigValidations`, `CantonEngineConfig`)
- handler wiring through command execution (`ApiServiceOwner`, `ApiServices`, `StoreBackedCommandInterpreter`) and node startup (`LedgerApiServer`, `ParticipantNode`)
- configuration, HTTP-handler and command-interpreter tests

## Out of scope

Recorded-result consistency checking and the extension-service validator implementation (PR 6), response routing (PR 7), and activation of validation in transaction processing (PR 8). The handler is dev-gated and off unless an extension service is configured, so the intermediate state is dormant.

## Stacking & review

This PR is stacked on PR 4, so its diff against `main` is **cumulative**. The incremental change for this PR alone (it depends on PR 3's handler SPI, not PR 4):
https://github.com/digital-asset/canton/compare/zenith-network:angelol/external-call-06-split-pr4-damle-mcc...zenith-network:angelol/external-call-06-split-pr5-extension

Refs #513.

PR #573, opened 2026-07-09
refactor(external-call): render ExtensionCallError via PrettyPrinting

angelol:
Addresses the deferred cleanup from #553 ([r3527995304](https://github.com/digital-asset/canton/pull/553#discussion_r3527995304)): make `ExtensionCallError` implement `PrettyPrinting` instead of interpolating its fields one by one, so a newly added field renders in the full-error log automatically rather than being silently dropped.

## Change

- `ExtensionCallError` now `extends PrettyPrinting` (`prettyOfClass`, with `paramIfDefined` for the optional `requestId`).
- The full-error log in `ExtensionServiceManager` renders `$error` via the pretty instead of listing each field.

The two ledger-api-client-facing renderings (`ExtensionServiceExternalCallHandler.sanitizedEngineError` and the `ExtensionServiceExternalCallValidator` failure message) are intentionally left field-by-field: they deliberately expose only the status code and request id, never the message, so they must not use the full pretty.

## Testing

`community-participant/Test/compile` green; `sbt format` clean. `ExtensionServiceExternalCallHandlerTest`, `ExtensionServiceExternalCallValidatorTest`, `HttpExtensionServiceClientTest` green — the four full-error log assertions were updated to the pretty format, and a case covering the optional `requestId` rendering was added.

### DIFF
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceClient.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceClient.scala
@@ -4,6 +4,7 @@
 package com.digitalasset.canton.participant.extension
 
 import com.digitalasset.canton.lifecycle.FutureUnlessShutdown
+import com.digitalasset.canton.logging.pretty.{Pretty, PrettyPrinting}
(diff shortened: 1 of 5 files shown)


matthiasS-da (approved):
LGTM

PR #574, opened 2026-07-09
fix(external-call): improve engine-extension config validation messages

angelol:
Addresses two deferred cleanups from #553 on the participant `engine.extensions` config validation messages ([r3527966232](https://github.com/digital-asset/canton/pull/553#discussion_r3527966232), [r3527974592](https://github.com/digital-asset/canton/pull/553#discussion_r3527974592)).

## Changes

- **Target-port message** now derives from `Port.Dynamic` instead of hardcoding `"between 1 and <maxValidPort>"`. The old wording was fragile (breaks if `Port.Dynamic` changes) and misleading — `Port.minValidPort == 0`, so `0` is a valid `Port` value and the range wrongly implied it was out of range. The check is `port == Port.Dynamic`, so the message now says the port must not be the dynamic port.
- **Retry-delay message** grammar: `respective values are (x, y)` → `respective values are x and y`.
- Adds the first tests for these two validations (there were none): dynamic port → error, fixed port → valid, and `retry-initial-delay > retry-max-delay` → error.

Scope note: the sibling `"respective values are (x, y, z)"` message in `ConfigValidations` belongs to the sequencer-client retry-delay validation (pre-existing, not external-call); it is intentionally left untouched here.

## Testing

`community-app/Test/compile` green; `sbt format` clean; `ConfigValidationsTest` green (the three new tests assert the exact corrected messages).

### DIFF
--- a/community/app/src/test/scala/com/digitalasset/canton/config/ConfigValidationsTest.scala
+++ b/community/app/src/test/scala/com/digitalasset/canton/config/ConfigValidationsTest.scala
@@ -643,6 +643,36 @@ class ConfigValidationsTest extends BaseTestWordSpec {
     }
   }
 
+  "engine extension api version checks" when {
+    "the version is not a valid URI path segment" should {
+      "fail" in {
+        assertErrors(
+          participantWithExtension(
+            ExtensionServiceConfig(
+              address = "localhost",
+              port = Port.tryCreate(8080),
+              version = "v1/extra",
+            )
+          )
+        )(
+          "For participant p1, engine.extensions.ext1.version must be a non-empty URI path" +
+            " segment containing only unreserved characters [A-Za-z0-9._~-], excluding '.' and" +
+            " '..', but found 'v1/extra'"
+        )
+      }
+    }
+
+    "a valid version is configured" should {
+      "succeed" in {
+        assertValid(
+          participantWithExtension(
+            ExtensionServiceConfig(address = "localhost", port = Port.tryCreate(8080))
+          )
+        )
+      }
+    }
+  }
+
   "nodes with crypto scheme configurations" should {
     def configWithSchemes(
         signingSchemeConfig: Option[SigningSchemeConfig] = None,
--- a/community/app/src/test/scala/com/digitalasset/canton/config/ConfigValidationsTest.scala
+++ b/community/app/src/test/scala/com/digitalasset/canton/config/ConfigValidationsTest.scala
@@ -21,6 +21,7 @@ import com.digitalasset.canton.http.JsonApiConfig
 import com.digitalasset.canton.integration.ConfigTransforms
 import com.digitalasset.canton.participant.config.{
   CantonEngineConfig,
+  ExtensionServiceConfig,
   LedgerApiServerConfig,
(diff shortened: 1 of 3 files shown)


community/app/src/test/scala/com/digitalasset/canton/config/ConfigValidationsTest.scala
```diff
@@ -588,6 +589,60 @@ class ConfigValidationsTest extends BaseTestWordSpec {
...
+              retryInitialDelay = NonNegativeFiniteDuration.ofSeconds(10),
+              retryMaxDelay = NonNegativeFiniteDuration.ofSeconds(5),
+            )
+          )
+        )(
+          "For participant p1, engine.extensions.ext1 retry delays must satisfy retry-initial-delay <= retry-max-delay; respective values are 10 seconds and 5 seconds"
+        )
+      }
+    }
+  }
```
matthiasS-da, line 644: By now, there is overlap with the tests you have created previously in CantonConfigTest. Could you please deduplicate that?
For me, it makes perfect sense to test config validations only here and not in CantonConfigTest.

angelol:
CI failure fixed in a9e768e25c: `CantonConfigTest`'s port-0 case asserted fragments of the old "between 1 and …" wording (`include("between 1")`), which this PR's reworded message no longer contains — the fragment form is also why no grep for the removed literal caught it. The assertion now pins the full new validation message. Verified locally: `CantonConfigTest` 48/48 green (the suite needs `POSTGRES_USER`/`POSTGRES_PASSWORD` set, which is why it wasn't in the local gate runs). The retry-delay case still passes unchanged — its asserted fragment survives the rewording.

community/app/src/test/scala/com/digitalasset/canton/config/ConfigValidationsTest.scala
angelol: Done in bfa7af1f8f: the port and retry-delay cases are removed from `CantonConfigTest` — they duplicated the `ConfigValidationsTest` coverage (the port one had also just broken CI by asserting fragments of the old wording, fixed in a9e768e25c before your comment; it is now gone entirely). The api-version case stays in `CantonConfigTest` for now since it has no `ConfigValidationsTest` counterpart — happy to move it over too if you'd like it consolidated.

matthiasS-da: > The api-version case stays in CantonConfigTest for now since it has no ConfigValidationsTest counterpart — happy to move it over too if you'd like it consolidated.

I happily accept that offer. Please migrate the tests over to ConfigValidationsTest.

angelol: Done in c7c6c84055: the api-version case now lives in `ConfigValidationsTest` (full message pinned, plus a passing counterpart), and `CantonConfigTest` no longer has a validation section for engine extensions. The one remaining extension test there exercises HOCON parsing (bearer-token auth mapping), not a validation rule, so it stayed.

matthiasS-da (approved):
LGTM

PR #575, opened 2026-07-09
test(external-call): fold prepare's await and annotate its return type

angelol:
Two deferred `ExternalTransactionProcessorSpec` cleanups from #553.

- **Fold `prepare`'s await** ([r3528203484](https://github.com/digital-asset/canton/pull/553#discussion_r3528203484)): `.value.failOnShutdown("prepare transaction").futureValue` moves into `prepare`, which now returns the awaited `Either[InteractiveSubmissionPreparationError.Reject, PrepareResult]`. Call sites reduce to `prepare(...)` plus `.value.hashVersion` / `.left.value.reason`.
- **Type annotation** ([r3528195602](https://github.com/digital-asset/canton/pull/553#discussion_r3528195602)): `prepare` was the only top-level helper `def` without a return-type annotation; it now has one.

Behavior-preserving; the four existing tests are unchanged in outcome.

The remaining `ExternalTransactionProcessorSpec` cleanup ([r3528190170](https://github.com/digital-asset/canton/pull/553#discussion_r3528190170) — simplify the parameter list via `testedProtocolVersion` + protocol-version gating) is a larger, semantics-affecting restructure and will follow as a separate PR on top of this one.

## Testing

`ledger-api-core/Test/compile` green; `sbt format` clean; `ExternalTransactionProcessorSpec` green (4/4).

Part of the #565 post-merge cleanup. Refs #565.

PR #576, opened 2026-07-10
test(external-call): parametrize ExternalTransactionProcessorSpec on testedProtocolVersion

angelol:
The remaining `ExternalTransactionProcessorSpec` cleanup from #553 ([r3528190170](https://github.com/digital-asset/canton/pull/553#discussion_r3528190170)): stop hardcoding protocol versions so each test runs once at the version it exercises, rather than redundantly at every supported version ([the pattern from #552 r3506896881](https://github.com/digital-asset/canton/pull/552#discussion_r3506896881)).

> **Builds on #575** (fold `prepare`'s await + annotate its return type). This PR's diff is cumulative until #575 merges; the incremental slice is [here](https://github.com/zenith-network/canton/compare/angelol/external-call-565-processor-spec-cleanup...angelol/external-call-565-processor-spec-pv-gating).

## Changes

- The test synchronizer now uses `testedProtocolVersion`; the two stable-behaviour tests are gated to `[v35, dev)` and the two VDev tests to `dev` (`onlyRunWithOrGreaterThan(ProtocolVersion.dev)`). The stable V4-rejection message is derived from `testedProtocolVersion`.
- Parameter list simplified per the review: `protocolVersion` gone (→ `testedProtocolVersion`), `submissionSeed` a fixed dummy, and `nodeSeeds` replaced by `nodeIds: Seq[NodeId]` with seeds generated internally.

The stable gate is `pv >= v35 && pv < dev` rather than a bare `< dev` because hashing scheme V3 is unsupported at v34 (a `< dev`-only gate would run the V3-success test at v34 and fail); a code comment records this.

## Testing

`ledger-api-core/Test/compile` green; `sbt format` clean. Ran the spec at both the default protocol version (v35 → the two stable tests run, the VDev tests are ignored) and `CANTON_PROTOCOL_VERSION=dev` (→ the VDev tests run, the stable tests are ignored) — all four pass at their gated version.

### DIFF
--- a/community/ledger/ledger-api-core/src/test/scala/com/digitalasset/canton/platform/apiserver/services/command/interactive/codec/ExternalTransactionProcessorSpec.scala
+++ b/community/ledger/ledger-api-core/src/test/scala/com/digitalasset/canton/platform/apiserver/services/command/interactive/codec/ExternalTransactionProcessorSpec.scala
@@ -27,7 +27,13 @@ import com.digitalasset.canton.platform.config.InteractiveSubmissionServiceConfi
 import com.digitalasset.canton.protocol.hash.HashTracer
 import com.digitalasset.canton.topology.DefaultTestIdentities
 import com.digitalasset.canton.version.{HashingSchemeVersion, ProtocolVersion}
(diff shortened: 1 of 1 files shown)


angelol:
Closing this — withdrawing to reconsider the approach.

PR #577, opened 2026-07-10
fix(external-call): forward the full error to the submitting client

angelol:
Revisits the submission-path error sanitization as agreed in [#553 r3511508412](https://github.com/digital-asset/canton/pull/553#discussion_r3511508412): the ledger-api-client-facing message now carries the full error instead of only a status code.

> Rebased onto `main` after #573 merged; the diff is now exactly this PR's change.

## What changes

`ExtensionServiceExternalCallHandler` used to reduce an `ExtensionCallError` to `"External call failed with status <code>, requestId=<id>"` before handing it to the engine. It now forwards the whole error via its pretty rendering, which the engine wraps with the call coordinates, so the rejection cause the submitting client sees becomes:

```
External call execution failed (extensionId=..., functionId=...): call failed: ExtensionCallError(status code = 503, message = "Service unavailable", request id = 'req-1', retryable = true)
```

The client gets the failure description, the request id for correlation with extension-service logs, and the retryability hint — and it is the same rendering the participant logs (#573), making log correlation trivial. The handler passes the bare rendered error (no prefix of its own) since the engine's rendering already provides the framing.

## Why forwarding the full message is safe (the "details" from the review thread)

- **Audience.** This path is reached only from `StoreBackedCommandInterpreter` (command interpretation on the submitting participant). The message surfaces in the `INTERPRETATION_EXTERNAL_CALL_ERROR_EXECUTION_FAILED` rejection to the authenticated submitting client — it is never distributed to other nodes (observation 4 in the thread). `IE.ExternalCall` is not catchable by Daml code, so the text cannot flow into committed ledger data.
- **Content.** Every `ExtensionCallError` construction site keeps `message` client-safe: `HttpExtensionServiceClient` maps HTTP statuses and exceptions to fixed strings ("Service unavailable", "Request timeout", …) and never copies response bodies into it (bodies go to logs only, via `diagnosticResponseBody`); the manager's unconfigured-extension error carries only the extension id the client itself supplied. The one non-fixed value any site interpolates is the configured max response-body size in the 413 message — a non-sensitive operational limit that is genuinely useful to the client (it says how big a response the extension may return). This content rule — fixed descriptions, identifiers, and non-sensitive operational limits; never response bodies or credentials — is now documented on `ExtensionCallError` so future construction sites keep the message client-safe.
- **The paths that face other nodes are untouched.** The Phase-3 validator (`ExtensionServiceExternalCallValidator`) still reduces the error to status code + request id before it can reach `LocalAbstain` confirmation responses, exactly as before (observation 2). The full-error log in `ExtensionServiceManager.handleExternalCall` from the original review also stays; only its now-stale "sanitization drops the message" comment is reworded.

Two consequences worth calling out explicitly:

- The auth-misconfiguration descriptions ("Failed to read bearer token file", "Bearer token file contains invalid characters", "Bearer token file is empty") now reach the submitting client. They are fixed strings — no paths, no token content — but they do reveal participant-side configuration state to the client; flagging in case you want that reduced further.
- The pretty rendering wraps at width 200, so a long error (e.g. the auth-config message plus a request id) can render across multiple lines inside the rejection cause — the same behavior the participant log already has. Happy to switch to a single-line rendering if preferred.

Not changed (and deliberately so): the error-code ids and the gRPC error category — retryability is conveyed in the message text, since a single `ErrorCode` cannot vary its category per error.

## Testing

`ExtensionServiceExternalCallHandlerTest` now asserts the full forwarded message; the previous `should not include "not configured"` no-leak assertions are reversed by design — forwarding that description to the submitting client is the point of this cleanup. Extension suites green at `CANTON_PROTOCOL_VERSION=dev` (handler, validator, HTTP client).

Part of the #565 post-merge cleanup (Part 2, item "revisit error sanitization"). Refs #565.

### DIFF
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceClient.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceClient.scala
@@ -10,18 +10,25 @@ import com.digitalasset.canton.tracing.TraceContext
 
 /** Error information from external call failures.
   *
-  * `message` must stay client-safe: fixed descriptions, identifiers, and non-sensitive operational
-  * limits only -- never extension-service response bodies or credentials. The submission path
-  * forwards the whole error to the submitting ledger API client in the command rejection, so
-  * anything placed here becomes client-visible. (The Phase-3 validation path reduces the error to
-  * status code and request id before it reaches confirmation responses.)
+  * `clientActionable` marks errors whose message is actionable feedback for the submitting ledger
+  * API client (unknown extension, oversized response, invalid header value): the submission path
+  * forwards their full rendering in the command rejection. All other errors concern the
+  * communication between the participant and the extension service, and the client receives only a
+  * generic message with the retry status. The full error is always logged by
+  * [[ExtensionServiceManager.handleExternalCall]]. `message` must stay client-safe regardless:
+  * fixed descriptions, identifiers, and non-sensitive operational limits only -- never
+  * extension-service response bodies or credentials. (The Phase-3 validation path reduces every
+  * error to status code and request id before it reaches confirmation responses.)
   */
 final case class ExtensionCallError(
     statusCode: Int,
     message: String,
     requestId: Option[String],
     retryable: Boolean,
+    clientActionable: Boolean,
 ) extends PrettyPrinting {
+  // `clientActionable` is deliberately not rendered: the rendering doubles as the client-facing
+  // payload for actionable errors, and the flag is control metadata rather than error content.
   override protected def pretty: Pretty[ExtensionCallError] = prettyOfClass(
     param("status code", _.statusCode),
     param("message", _.message.doubleQuoted),
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceClient.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceClient.scala
@@ -14,26 +14,32 @@ import com.digitalasset.canton.tracing.TraceContext
   * API client (unknown extension, oversized response, invalid header value): the submission path
   * forwards their full rendering in the command rejection. All other errors concern the
   * communication between the participant and the extension service, and the client receives only a
-  * generic message with the retry status. The full error is always logged by
-  * [[ExtensionServiceManager.handleExternalCall]]. `message` must stay client-safe regardless:
-  * fixed descriptions, identifiers, and non-sensitive operational limits only -- never
+  * generic message with the retry status and the tracing identifiers. The full error is always
+  * logged by [[ExtensionServiceManager.handleExternalCall]]. `message` must stay client-safe
+  * regardless: fixed descriptions, identifiers, and non-sensitive operational limits only -- never
   * extension-service response bodies or credentials. (The Phase-3 validation path reduces every
-  * error to status code and request id before it reaches confirmation responses.)
+  * error to status code and external call id before it reaches confirmation responses.)
+  *
+  * `externalCallId` is minted per HTTP request by [[HttpExtensionServiceClient]] (sent to the
+  * service as `X-Request-Id`); it is named to disambiguate from the Canton request id. Together
+  * with the trace id it correlates an error with the participant's and the service's logs.
   */
 final case class ExtensionCallError(
     statusCode: Int,
     message: String,
-    requestId: Option[String],
+    externalCallId: Option[String],
     retryable: Boolean,
     clientActionable: Boolean,
-) extends PrettyPrinting {
+)(implicit val traceContext: TraceContext)
+    extends PrettyPrinting {
   // `clientActionable` is deliberately not rendered: the rendering doubles as the client-facing
   // payload for actionable errors, and the flag is control metadata rather than error content.
   override protected def pretty: Pretty[ExtensionCallError] = prettyOfClass(
     param("status code", _.statusCode),
     param("message", _.message.doubleQuoted),
-    paramIfDefined("request id", _.requestId.map(_.singleQuoted)),
+    paramIfDefined("external call id", _.externalCallId.map(_.singleQuoted)),
     param("retryable", _.retryable),
+    paramIfDefined("trace id", _.traceContext.traceId.map(_.singleQuoted)),
   )
 }
 
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceClient.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceClient.scala
@@ -8,7 +8,14 @@ import com.digitalasset.canton.logging.pretty.{Pretty, PrettyPrinting}
 import com.digitalasset.canton.platform.execution.ExternalCallMode
 import com.digitalasset.canton.tracing.TraceContext
 
-/** Error information from external call failures. */
+/** Error information from external call failures.
+  *
+  * `message` must stay client-safe: fixed descriptions, identifiers, and non-sensitive operational
+  * limits only -- never extension-service response bodies or credentials. The submission path
+  * forwards the whole error to the submitting ledger API client in the command rejection, so
+  * anything placed here becomes client-visible. (The Phase-3 validation path reduces the error to
+  * status code and request id before it reaches confirmation responses.)
+  */
 final case class ExtensionCallError(
     statusCode: Int,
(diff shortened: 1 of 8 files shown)


matthiasS-da:
For discussion, here is a summary of the information leaked through error messages to ledger api clients:

1. s"Extension '$extensionId' not configured"
2. Status code from Http response (converted to a human readable message such as “Bad Request”, “Rate limit exceeded”)
3. s"External call response body exceeded maximum size of ${tooLarge.maxResponseBodyBytes} bytes"
4. HttpTimeoutException => "Request timeout"
5. ConnectException “=> Connection failed"
6. IOException => "I/O error"
7. Throwable => “Unexpected error”
8. s"Invalid external-call HTTP header value for ${X-Daml-External-Function-Id|X-Daml-External-Config-Hash}”, if function id or config hash are too long or contain invalid characters.
9. s"Invalid extension service authentication configuration: $error" with error being one of:
    * "Failed to read bearer token file"
    * "Bearer token file contains invalid characters"
    * "Bearer token file is empty"

They fall into the following categories:
While some (1, 3, 8) are probably ok to leak, others concern communication problems between participant and external service, including participant misconfiguration.

We'll discuss internally and get back to you.

@soren-da FYI

matthiasS-da:
> For discussion, here is a summary of the information leaked through error messages to ledger api clients:
> 
> 1. s"Extension '$extensionId' not configured"
> 2. Status code from Http response (converted to a human readable message such as “Bad Request”, “Rate limit exceeded”)
> 3. s"External call response body exceeded maximum size of ${tooLarge.maxResponseBodyBytes} bytes"
> 4. HttpTimeoutException => "Request timeout"
> 5. ConnectException “=> Connection failed"
> 6. IOException => "I/O error"
> 7. Throwable => “Unexpected error”
> 8. s"Invalid external-call HTTP header value for ${X-Daml-External-Function-Id|X-Daml-External-Config-Hash}”, if function id or config hash are too long or contain invalid characters.
> 9. s"Invalid extension service authentication configuration: $error" with error being one of:
>    
>    * "Failed to read bearer token file"
>    * "Bearer token file contains invalid characters"
>    * "Bearer token file is empty"
> 
> They fall into the following categories: While some (1, 3, 8) are probably ok to leak, others concern communication problems between participant and external service, including participant misconfiguration.
> 
> We'll discuss internally and get back to you.
> 
> @soren-da FYI

@angelol While discussing it internally, we got the impression that cases 1, 3, 8 contain actionable feedback for ledger api clients. In the other cases, there are communication problems between participant and external service and consequently the error message is not actionable for clients. Clients only need to know the retry status.
Consequently:

- Please let us know if you agree with our assessment and why / why not.
- If you agree, could you please replace the messages sent to clients in cases other than 1, 3, 8 by the same generic message. (The original message should still be logged server-side.) This is to avoid unnecessary information leakage while still exposing details when they are useful to honest clients.

angelol:
Agreed with the assessment — for cases other than 1/3/8 the client cannot act on the details (the failure sits between the participant and the extension service, and only the operator can fix a timeout, an auth misconfiguration, or a 5xx), so exposing them adds leakage without helping an honest client.

Implemented in 2e6832bdf9. `ExtensionCallError` now carries an explicit `clientActionable` flag set at each construction site (1/3/8 true, everything else false); the submission handler forwards the full rendering only for actionable errors, and everything else becomes the single generic message. The full error is still logged by `ExtensionServiceManager` as before, and the Phase-3 confirmation-response path is unchanged.

One deliberate addition beyond the letter of the proposal: the generic message keeps the participant-minted request id next to the retry status — `External call failed (retryable = true, request id = '…')` — since it is a random UUID that reveals nothing about the failure but lets a client hand the operator a correlation handle for the server-side log. Happy to drop it if you prefer the strictly minimal form.

community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceClient.scala
```diff
@@ -8,13 +8,27 @@ import com.digitalasset.canton.logging.pretty.{Pretty, PrettyPrinting}
...
+  * generic message with the retry status. The full error is always logged by
+  * [[ExtensionServiceManager.handleExternalCall]]. `message` must stay client-safe regardless:
+  * fixed descriptions, identifiers, and non-sensitive operational limits only -- never
+  * extension-service response bodies or credentials. (The Phase-3 validation path reduces every
+  * error to status code and request id before it reaches confirmation responses.)
+  */
 final case class ExtensionCallError(
     statusCode: Int,
     message: String,
     requestId: Option[String],
```
matthiasS-da: Looking at this again, I notice that this is a new kind of request id, introduced in HttpExtensionServiceClient. We already have several ids for tracing:
- The Canton requestId - a timestamp, but not available during submission.
- The command id - uniquely identifies the ledger command, but not available for validating participants.
- The TraceContext - uniquely identifies a request, also for validating participants

For best possible tracing of errors, I suggest to include the trace id and the request id from the HttpExtensionServiceClient. I kindly request the following modifications: 
- Please include the TraceContext as a parameter of this class.
- Please rename requestId to disambiguate from the Canton requestId, e.g., to externalCallId.

angelol: Done in b5e27b99a1. `ExtensionCallError` now carries the `TraceContext` in a second implicit parameter list (the `Traced` pattern, keeping case-class equality on the payload fields), and the field is renamed `externalCallId`. Both identifiers render in the error's pretty form — and therefore in the full message actionable errors forward to the client — and in the generic client message for the non-actionable cases: `External call failed (retryable = ..., external call id = '...', trace id = '...')`.

Included for consistency, since they are the same identifier: the internal locals and log labels in `HttpExtensionServiceClient` and the Phase-3 validator's reduced reason (`externalCallId=` instead of `requestId=`) follow the rename. The `X-Request-Id` header sent to the extension service is unchanged. One scope note: the Phase-3 reduction gains only the rename, not the trace id — confirmation-response processing already carries the request's trace context — but happy to add it there too if you'd like.

community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceClient.scala
```diff
@@ -8,18 +8,38 @@ import com.digitalasset.canton.logging.pretty.{Pretty, PrettyPrinting}
...
+    extends PrettyPrinting {
+  // `clientActionable` is deliberately not rendered: the rendering doubles as the client-facing
+  // payload for actionable errors, and the flag is control metadata rather than error content.
   override protected def pretty: Pretty[ExtensionCallError] = prettyOfClass(
     param("status code", _.statusCode),
     param("message", _.message.doubleQuoted),
-    paramIfDefined("request id", _.requestId.map(_.singleQuoted)),
+    paramIfDefined("external call id", _.externalCallId.map(_.singleQuoted)),
     param("retryable", _.retryable),
+    paramIfDefined("trace id", _.traceContext.traceId.map(_.singleQuoted)),
```
matthiasS-da: Please simplify this to `param("trace id", _.traceContext),`

matthiasS-da (commented):
LGTM except for one small comment. Once this is addressed, I'll approve and merge.

matthiasS-da:
@paulbrauner-da I think we are [blocked](https://app.circleci.com/pipelines/github/digital-asset/canton/256/workflows/b04d4de6-5501-494c-a244-104ca4e0063e/jobs/4528) by an outdated main branch. Could you please do a code-drop from the private repo so we can continue here? Please make sure to preserve PRs that have been merged to the public repo since the last code drop.

@thibault-da: Paul is ooo, could you please take over?

community/participant/src/main/scala/com/digitalasset/canton/participant/extension/ExtensionServiceClient.scala
angelol: Done in 250f97c3c — `param("trace id", _.traceContext)`, delegating to `Pretty[TraceContext]` (`tid:<traceId>`, bare `tid:` for an empty context, now always rendered). Log and message assertions updated accordingly.

matthiasS-da (approved):
LGTM - all comments addressed

matthiasS-da:

angelol:
Rebased onto latest main (eef8fc040); new tip 9a249b65ed7812f4b2228887ce07a3c8d275ccc2. All commits are unchanged from the reviewed state (range-diff clean); the full participant suite is green locally at both stable and dev protocol versions.

PR #578, opened 2026-07-10
test(external-call): simplify ExternalTransactionProcessorSpec parameters

angelol:
Simplifies `ExternalTransactionProcessorSpec`'s helper parameter list as requested in [#553 r3528190170](https://github.com/digital-asset/canton/pull/553#discussion_r3528190170), following the `testedProtocolVersion` approach from [#552 r3506896881](https://github.com/digital-asset/canton/pull/552#discussion_r3506896881).

> Rebased onto `main` after #575 merged; the diff is now exactly this PR's change.

## What changes

- **`protocolVersion`**: dropped from `commandExecutionResultFor`/`prepare`; the helpers use the tested protocol version (the synchronizer id now comes straight from `DefaultTestIdentities.physicalSynchronizerId`, which already carries it), and each test runs on the protocol-version range it is actually about (`ProtocolVersionChecksAnyWordSpec`):
  - "honor the requested hashing scheme": requests `testedHashingSchemeVersion` (the highest scheme supported at the tested version), so it runs meaningfully at every protocol version — including v34, which the old v35-hardcoded test never covered — and the "non-dev" qualifier goes away.
  - both VDev-transaction tests: `onlyRunWithOrGreaterThan(dev)` — a VDev transaction is a dev-PV artifact, and `processPrepare` encodes before the hash check, so running them at stable versions would exercise the encoder with input that cannot occur there.
  - "reject V4 below dev": `onlyRunWhen(_ < dev)` — the below-dev counterpart of the V4-honored-at-dev case ("below dev" rather than "stable", since the gate also covers alpha versions). Its expected message now interpolates `testedProtocolVersion` and derives the supported-scheme list from `HashingSchemeVersion.getHashingSchemeVersionsForProtocolVersion`, so it stays exact at v34 ("V2") as well as v35 ("V2, V3").
- **`submissionSeed`**: dropped; the helper uses a fixed dummy (the per-test seed strings had no assertion value).
- **`nodeSeeds`**: `prepare` now takes only the node ids and generates the seeds itself.

`vDevCreateTransaction`'s contract-id seed parameter is left as is — the per-test strings there keep contract ids distinct across tests, which is orthogonal to this comment.

## Testing

Suite exercised at both `CANTON_PROTOCOL_VERSION=dev` (V3 + both VDev tests run, V4-on-stable ignored) and the default stable version (V3 + V4-on-stable run, VDev tests ignored) — green in both, with the ignored counts matching the gates.

Part of the #565 post-merge cleanup (Part 2). Refs #565.

### DIFF
--- a/community/ledger/ledger-api-core/src/test/scala/com/digitalasset/canton/platform/apiserver/services/command/interactive/codec/ExternalTransactionProcessorSpec.scala
+++ b/community/ledger/ledger-api-core/src/test/scala/com/digitalasset/canton/platform/apiserver/services/command/interactive/codec/ExternalTransactionProcessorSpec.scala
@@ -27,7 +27,13 @@ import com.digitalasset.canton.platform.config.InteractiveSubmissionServiceConfi
 import com.digitalasset.canton.protocol.hash.HashTracer
 import com.digitalasset.canton.topology.DefaultTestIdentities
 import com.digitalasset.canton.version.{HashingSchemeVersion, ProtocolVersion}
-import com.digitalasset.canton.{BaseTest, HasExecutionContext, LedgerUserId, LfTimestamp}
+import com.digitalasset.canton.{
+  BaseTest,
+  HasExecutionContext,
+  LedgerUserId,
+  LfTimestamp,
+  ProtocolVersionChecksAnyWordSpec,
+}
 import com.digitalasset.daml.lf.command.ApiCommands
 import com.digitalasset.daml.lf.crypto.Hash as LfHash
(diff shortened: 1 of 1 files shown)


matthiasS-da (dismissed):
LGTM

matthiasS-da:
@angelol Please resolve the merge conflicts.

angelol:
Rebased onto `main` (the branch previously carried #575's commit, which conflicted after its squash-merge) — conflicts are resolved and the diff is now exactly this PR's change. Suite re-verified at dev and stable protocol versions.

matthiasS-da (approved):
LGTM

PR #581, opened 2026-07-15
refactor(external-call)!: remove the per-party disagreement machinery

angelol:
Implements the first action of the [#565 decision](https://github.com/digital-asset/canton/issues/565#issuecomment-4938124738 ): per-party rejection routing is not needed, so the machinery supporting it goes away.

## What changes

Removes `ExternalCallConsistencyChecker.Result#hostedInconsistencies` and everything that existed only to feed it: the derived `inconsistentParties`, the private per-party grouping, the `hostedConfirmingParties` parameter of `check`, and the `checkingParties` copy in the checker-private `VisibleExternalCallOccurrence`. This is a pure deletion — the path was already dead in production, since the only caller passed `hostedConfirmingParties = Set.empty`.

One knock-on: with a single field left, the checker's `Result` wrapper serves no purpose, so `check` now returns the visible inconsistencies directly.

Deliberately kept: `ViewParticipantData.ViewExternalCallResult.checkingParties`. It is a serialized wire field, protected by model-conformance reconstruction, and the input to the upcoming re-validation restriction (point 3 of the decision, follow-up PR).

## Testing

`ExternalCallConsistencyCheckerTest` rewritten to the slimmed API: the per-party routing tests collapse into visible-inconsistency tests, and the fixture test for a conflicting-outputs-within-one-view scenario is deleted — since the fail-fast in `TransactionView.validated` (#572), such a view cannot be deserialized, so the scenario is unreachable in production. Suites green at `CANTON_PROTOCOL_VERSION=dev`; whole-repo `scalafixCheck` clean.

First of three PRs for the #565 per-view rework (removal → per-view results → re-validation restriction). Refs #565.

### DIFF
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallConsistencyChecker.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallConsistencyChecker.scala
@@ -124,20 +124,15 @@ object ExternalCallConsistencyChecker {
         }
       }
 
-  private def visibleInconsistencies(
-      occurrences: Seq[VisibleExternalCallOccurrence]
-  ): Seq[Inconsistency] =
-    occurrences
+  /** Returns the disagreements across all occurrences visible to this participant, sorted
+    * deterministically (by key, then outputs, then occurrences).
+    */
+  def check(views: Map[ViewPosition, ParticipantTransactionView]): Seq[Inconsistency] =
+    visibleOccurrences(views)
       .groupMap(_.key)(occurrence => occurrence.output -> occurrence.occurrence)
       .toSeq
       .flatMap { case (key, outputsAndOccurrences) =>
         inconsistencyFor(key, outputsAndOccurrences)
       }
       .sorted(orderInconsistency)
-
-  /** Returns the disagreements across all occurrences visible to this participant, sorted
-    * deterministically (by key, then outputs, then occurrences).
-    */
-  def check(views: Map[ViewPosition, ParticipantTransactionView]): Seq[Inconsistency] =
-    visibleInconsistencies(visibleOccurrences(views))
 }
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallConsistencyChecker.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallConsistencyChecker.scala
@@ -4,7 +4,6 @@
 package com.digitalasset.canton.participant.protocol.validation
 
 import cats.Order
-import com.digitalasset.canton.LfPartyId
 import com.digitalasset.canton.config.RequireTypes.NonNegativeInt
 import com.digitalasset.canton.data.ExternalCallPayloadDescription.{byteCount, hexPayloadSize}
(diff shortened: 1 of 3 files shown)


community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallConsistencyChecker.scala
```diff
@@ -174,38 +135,9 @@ object ExternalCallConsistencyChecker {
...
-      hostedConfirmingParties: Set[LfPartyId],
-  ): Result = {
-    val occurrences = visibleOccurrences(views)
-    Result(
-      hostedInconsistencies = hostedInconsistencies(occurrences, hostedConfirmingParties),
-      visibleInconsistencies = visibleInconsistencies(occurrences),
-    )
-  }
+  def check(views: Map[ViewPosition, ParticipantTransactionView]): Seq[Inconsistency] =
+    visibleInconsistencies(visibleOccurrences(views))
```
matthiasS-da: MINOR: Let's please inline visibleInconsistencies, as this is the only usage.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallConsistencyCheckerTest.scala
```diff
@@ -63,108 +57,36 @@ class ExternalCallConsistencyCheckerTest
...
-      val visibleInconsistency = result.visibleInconsistencies.loneElement
-      visibleInconsistency.outputs shouldBe Set(
+      inconsistency.outputs shouldBe Set(
         externalCallResult.output,
         otherExternalCallResult.output,
       )
+      inconsistency.occurrences.map(_.viewPosition) shouldBe Set(
+        ViewPosition.root,
+        ViewPosition(
+          List(ViewPosition.MerkleSeqIndex(List(ViewPosition.MerkleSeqIndex.Direction.Right)))
```
matthiasS-da: MINOR: Please define and use constants `leftPosition` and `rightPosition` to refer to these view positions.

matthiasS-da (commented):
LGTM except for a few minor comments

community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallConsistencyChecker.scala
angelol: Done in b4091c834.

community/participant/src/test/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallConsistencyCheckerTest.scala
angelol: Done in b4091c834 (`leftPosition`/`rightPosition`, used at all former literal sites).

matthiasS-da (approved):
LGTM - all comments addressed

matthiasS-da:

angelol:
Rebased onto latest main (eef8fc040); new tip c9bd818e7a3b333e57675ef9d3d71e9f51bdd681. Commits unchanged from the reviewed state (range-diff clean); participant suite green locally at both stable and dev protocol versions. #582/#583 have been re-stacked on top.

PR #582, opened 2026-07-15
feat(external-call)!: produce and consume external-call check results per view

angelol:
Implements point 4 of the [#565 decision](https://github.com/digital-asset/canton/issues/565#issuecomment-4938124738 ): external-call validation is computed and consumed per view.

> **Update (2026-07-22):** #581 has merged and this PR has been rebased onto latest main — the diff is now exactly this PR's single commit.

## What changes

- `ExternalCallCheck.check` returns one result per view that records external-call results (`Map[ViewPosition, Result]`); views without external calls have no entry.
- `TransactionConfirmationResponsesFactory` consults only the result of the view it responds for, following the existing per-position pattern of the authentication and authorization results. This fixes the two defects named in the decision — rejections duplicated across responses, and rejections attached to views without external calls — plus a third of the same family: the aggregate `Rejected` carried the globally-first inconsistency's description, so a view could be rejected with a message about calls recorded in other views. Each view's rejection now describes a call recorded in that view.
- Attachment rules per the decision: a re-validation failure rejects exactly the views whose participant data records the failed call; a disagreement rejects exactly the views that record the disagreeing key. Within a view, a disagreement takes precedence over a mismatch, which takes precedence over an unvalidatable result — mirroring the previous request-level precedence.
- Re-validation suppression is now per key: keys whose visible occurrences disagree are still not re-validated (there is no unambiguous output to validate), but other keys now are — previously any disagreement suppressed all re-validation, which under per-view results would have masked an independent re-validation failure in an unaffected view.
- Alarms are unchanged: every disagreement and every mismatch is alarmed once, at request level, so the per-view fan-out does not multiply alarms.

## Testing

`ExternalCallProtocolIntegrationTest` rewritten for per-view semantics; the response-factory cases now attach results to specific views (the previous fixtures injected verdicts into views without any external calls — an executable specification of the defect this PR removes) and assert reject/abstain on exactly the affected view with approval of the other. New coverage: a mismatch on a key recorded in several views rejects all of them after a single validator invocation; a re-validation failure is isolated to the recording view; a disagreement no longer suppresses re-validation of unaffected calls. Suites green at `CANTON_PROTOCOL_VERSION=dev`; whole-repo `scalafixCheck` clean.

Second of three PRs for the #565 per-view rework. Refs #565.

### DIFF
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessingSteps.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/TransactionProcessingSteps.scala
@@ -1684,7 +1684,7 @@ object TransactionProcessingSteps {
         ErrorWithInternalConsistencyCheck,
         Unit,
       ],
-      externalCallCheckResultF: FutureUnlessShutdown[ExternalCallCheck.Result],
+      externalCallCheckResultF: FutureUnlessShutdown[Map[ViewPosition, ExternalCallCheck.Result]],
       timeValidationResultE: Either[TimeCheckFailure, Unit],
       replayCheckResult: Option[String],
   ) {
--- a/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallCheck.scala
+++ b/community/participant/src/main/scala/com/digitalasset/canton/participant/protocol/validation/ExternalCallCheck.scala
@@ -3,6 +3,7 @@
 
 package com.digitalasset.canton.participant.protocol.validation
 
+import cats.syntax.functor.*
 import com.digitalasset.canton.config.RequireTypes.PositiveInt
 import com.digitalasset.canton.data.{
   ExternalCallKey,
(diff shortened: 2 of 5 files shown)


matthiasS-da (dismissed):
LGTM

matthiasS-da:

angelol:
Re-stacked onto the rebased #581 — its review-fix commit now sits underneath, as it should. New tip d83ca4beac1b704c6771fff103d9a6b79ad5dedc; this PR's own commit is unchanged from the approved state (range-diff clean); participant suite green locally at both stable and dev protocol versions.

angelol:
Rebased past the merged #581 (and #577) onto latest main — the conflict is resolved and GitHub reports the PR as mergeable again. New tip 1b407aebc2abbed628e71621b62e86d8d72a1dba; the commit is unchanged from the approved state (range-diff clean). #583 has been re-stacked on top (new tip 0cad953f54e03e9367cf96a816528319b225b310). Participant suite green locally at both stable and dev protocol versions on both tips.

matthiasS-da (approved):
LGTM

PR #591, opened 2026-07-23
test(external-call): pin the negative-path failure logs and client messages

angelol:
Second slice of #564: the unit-level half of the negative-path test work requested at the [#553 approval](https://github.com/digital-asset/canton/pull/553#issuecomment-4891915028) — "these tests should check failures reported to ledger api clients as well as log messages". (The other half — asserting the gRPC rejection a ledger-API client observes — needs the mock-extension-service integration fixture and comes with the e2e test PR.)

## What changes

**`HttpExtensionServiceClientTest`** — the failure-path tests suppressed warnings wholesale and asserted message fragments; they now pin the behavior:
- The exception paths (oversized body, stalled-body timeout, connection failure) assert the full warning message — up to the random external call id, hence a regex with the UUID pattern pinned — and that the response body does not leak into the log.
- Every other negative path (HTTP status mapping, header validation, bearer-token preflight, the manager's 404) emits no warning at all: status errors log the diagnostic body at INFO, and finite-`maxRetries` retries log each attempt and the give-up at INFO. Their blanket suppressions were vestigial and are dropped. The shutdown test's suppression stays — closing mid-retry warns nondeterministically.
- Client-visible messages upgrade from fragment to full-message asserts throughout; where full equality already proves an identifier cannot leak (token, token file path, sibling extension ids, rejected header values), the redundant negative checks are folded into it.
- The startup tests assert the exact preflight failure and pin the remote-validation failure's structure via regex (its tail is the JDK's connection-refused message).

**`ExtensionServiceExternalCallHandlerTest`** — the non-actionable-error test stubbed the manager method that contains the full-error warning, so the log was untestable. It now runs against a real `ExtensionServiceManager` over a local HTTP server returning 503 and asserts the pair of observables together: the warning carries the full internal error while the engine receives only the generic message. The `Unsupported` path's assertion becomes a full-message check; it emits no log by design (no logger there), so no log assertion is added — doing so would need a production change, deliberately out of scope for a test-only PR.

One commit per suite. Locally green: the three extension suites 39/0 at dev PV, `sbt format`, whole-repo `scalafixCheck`.

### DIFF
--- a/community/participant/src/test/scala/com/digitalasset/canton/participant/extension/ExtensionServiceExternalCallHandlerTest.scala
+++ b/community/participant/src/test/scala/com/digitalasset/canton/participant/extension/ExtensionServiceExternalCallHandlerTest.scala
@@ -4,11 +4,16 @@
 package com.digitalasset.canton.participant.extension
 
 import com.digitalasset.canton.config.ProcessingTimeout
-import com.digitalasset.canton.lifecycle.FutureUnlessShutdown
+import com.digitalasset.canton.config.RequireTypes.{NonNegativeInt, Port}
+import com.digitalasset.canton.participant.config.ExtensionServiceConfig
 import com.digitalasset.canton.platform.execution.ExternalCallMode
 import com.digitalasset.canton.tracing.TraceContext
 import com.digitalasset.canton.{BaseTest, HasExecutionContext}
+import com.sun.net.httpserver.{HttpHandler, HttpServer}
 import org.scalatest.wordspec.AnyWordSpec
+
+import java.net.InetSocketAddress
+import java.nio.charset.StandardCharsets
 
 class ExtensionServiceExternalCallHandlerTest
     extends AnyWordSpec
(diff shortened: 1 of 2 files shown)


simonmaxen-da:
LGTM

simonmaxen-da (approved):

angelol:
Rebased onto current `main` (c757b4922) and pushed — commits content-identical to the approved ones (`git range-diff` shows both as `=`), new tip 8e57b63f8. Touched suites re-run green locally at both the stable default and `CANTON_PROTOCOL_VERSION=dev`.

angelol:
Rebased onto current main c2481c28c per the Slack request — new tip 97bee77e1, range-diff against the approved 8e57b63f8 is all `=` (content unchanged, only the base moved). Re-verified locally: the three extension suites 39/0 at both dev and stable PV; format + scalafixCheck clean.

angelol:
Rebased onto main 86bad0fd1. New tip 800e080de, content unchanged (range-diff clean). Extension suites 39/0 at dev and stable, format and scalafixCheck clean.

PR #599, opened 2026-08-03
Fail indexer batch on unresolved deactivation reference

### LINKED ISSUE #598: Indexer writes deactivation events with NULL activation back-pointer on resolution failure, permanently corrupting the Ledger API ACS
## Summary

The parallel indexer computes the `deactivated_event_sequential_id` back-pointer
for each deactivation row at ingestion time. When resolution fails, it logs a
WARN and writes the row anyway, with `deactivated_event_sequential_id = NULL`
and `internal_contract_id = NULL`
(`ParallelIndexerSubscription.refillMissingDeactivatedActivations`).

Every ACS activeness query anti-joins on exactly that back-pointer
(`UpdateStreamingQueries.eventNotDeactivated`), not on the contract id. A
deactivation row with a NULL pointer therefore never matches: the archive is
visible in the update/transaction stream, but the contract stays ACTIVE in
every Ledger API ACS read, permanently. Nothing re-derives the pointer later.

We confirmed this in production on a Canton Network mainnet participant
(version `3.5.0-snapshot.20260401.14638.0.v9a1531c5`): 3 archived contracts
served as active for 7+ weeks. Their deactivate rows exist with `event_type = 3`
and NULL back-pointer; each has exactly one activation row; the sync-side ACS
(`par_active_contracts`) archived them correctly, so ACS commitments never
flagged anything. We found the divergence only because independent operators
compared aggregate values computed from their respective ACS reads. Full
incident details, ledger identifiers, and diagnostics are in DA support ticket
DA-4793.

## Why this is severe

- **Silent**: the only signal is a WARN
  (`"Activation is missing for a deactivation for ..."` /
  `"internal_contract_id for contract ID: ... not found, deactivation
  reference cannot be computed"`). Ingestion continues.
- **Permanent**: no repair or backfill path exists; the corruption is stable
  across restarts.
- **Invisible to the integrity checker**: the stray-deactivations check in
  `IntegrityStorageBackendImpl` only inspects rows where the back-pointer is
  NOT NULL.
- **Made worse by pruning**: the pruning-candidate join goes through the
  back-pointer, so pruning deletes the orphaned deactivate row but can never
  delete the phantom activation it should have killed. After pruning, the
  corruption is no longer even diagnosable from the DB.
- **Divergent reads across nodes**: participants hosting the same
  (multi-hosted) party serve different ACS contents. For token registries this
  means different total-supply answers per node. Commitments do not catch it
  because they are computed from the (correct) sync-side ACS.

## Code path (paths relative to the repository root)

- Deactivate DTO emitted with `deactivated_event_sequential_id = None,
  // this is filled later`: `community/ledger/ledger-api-core/src/main/scala/com/digitalasset/canton/platform/store/backend/UpdateToDbDto.scala`
- In-memory resolution (`seqMapper` / `tryToGetDeactivated`) and DB fallback
  (`dbPrepare`: `lookupBatchedInternalIdsNonReadThrough` on `par_contracts`,
  then `lastActivations` bounded by the ledger-end cache):
  `community/ledger/ledger-api-core/src/main/scala/com/digitalasset/canton/platform/indexer/parallel/ParallelIndexerSubscription.scala`
- The silent failure: `refillMissingDeactivatedActivations` →
  `dbDto.withActivationRef(None)` on `Some(None)`, WARN only (same file)
- The activeness predicate that can then never match:
  `community/ledger/ledger-api-core/src/main/scala/com/digitalasset/canton/platform/store/backend/common/UpdateStreamingQueries.scala`
  (`eventNotDeactivated`)
- Integrity check that skips NULL pointers:
  `community/ledger/ledger-api-core/src/main/scala/com/digitalasset/canton/platform/store/backend/common/IntegrityStorageBackendImpl.scala`

## Reproduction / detection

We do not yet have a minimal reproduction of the resolution failure itself
(in our incident, the same lookups succeed when run today, so the trigger was
a transient condition — candidates: indexer restart/replay interleaving, the
contract-store pruning race that already has a restart-recovery path via
`ReferencedContractNotFoundException`, or a ledger-end-cache window).

Detection on any affected participant is trivial:

    SELECT count(*) FROM lapi_events_deactivate_contract
    WHERE deactivated_event_sequential_id IS NULL;

plus a log grep for "Activation is missing for a deactivation" /
"deactivation reference cannot be computed".

## Proposed fix (implemented in #599)

1. **Fail the ingestion batch instead of writing NULL.** When
   `refillMissingDeactivatedActivations` (or the preceding internal-id
   resolution) cannot resolve an activation for a deactivation, throw and let
   the indexer restart — the same recovery pattern the pruning race already
   uses (`ReferencedContractNotFoundException`), including evicting the
   involved contracts from the contract cache so a stale cache mapping cannot
   cause a restart loop. On restart, resolution runs against settled state; in
   our production incident that retry would have succeeded. A crash-loop on a
   persistent failure is a visible incident; the current behavior is silent
   permanent corruption.
2. **Extend the integrity checker** to flag deactivation rows below the ledger
   end whose back-pointer is NULL, so affected deployments can detect existing
   corruption before pruning hides it.
3. A counter metric for these restarts so operators can alert on it.

A more thorough alternative would be to make the resolution itself robust
(e.g., a fallback lookup by contract id not bounded by the ledger-end cache).
We did not attempt this because the cache bound looks like a deliberate
consistency invariant — maintainer guidance welcome on whether that route is
preferable. Also welcome: confirmation whether any legitimate flow (e.g. ACS
import during party replication) can produce an unresolvable deactivation —
if so, the fail-fast needs a carve-out for it.

## Environment

- Canton Network mainnet (global synchronizer), participant
  `3.5.0-snapshot.20260401.14638.0.v9a1531c5`
- Affects Canton Network / splice deployments via the
  `com.daml:ledger-api-core` binary dependency (currently 3.5.11)
- Incident window: 2026-06-11; discovered 2026-07-30 via cross-operator
  ACS comparison
- Blast radius on the affected node: exactly 3 deactivation rows with NULL
  back-pointer (query above)

scolear:
Fixes #598.

## What

The indexer computes the `deactivated_event_sequential_id` back-pointer for each deactivation event at ingestion time. When resolution failed, it logged a warning and persisted the event with a NULL back-pointer. Every Ledger API activeness query anti-joins on that back-pointer, so the archived contract stayed visible as active in the ACS permanently. We observed this in production: a mainnet participant served 3 archived contracts as active for 7 weeks, diverging from every other participant hosting the same party (details in #598 and DA support ticket DA-4793).

This PR makes the failure loud instead of corrupting:

1. `dbPrepare` now fails the ingestion batch when any deactivation's activation reference stays unresolved, following the `ReferencedContractNotFoundException` recovery pattern: the indexer restarts and retries. Before throwing, it invalidates the cached contract lookups for the affected contracts, so a stale cached mapping cannot cause a restart loop.
2. The tolerant branch in `refillMissingDeactivatedActivations` becomes an invariant violation, closing the only code path that wrote NULL back-pointers.
3. A new integrity check reports deactivation events with a NULL activation reference at or above the pruning point and below the ledger end. The existing checks either skipped NULL back-pointers or treated them as pruning leftovers.
4. A new `Errors`-qualified counter, `indexer_restart_due_to_unresolved_deactivation`, makes the restarts alertable.

## Design decisions

**Cache invalidation uses a new `ContractStore.invalidateCachedContracts` method** instead of the existing `contractsPruned`. The existing method deletes contract data on the in-memory store (correct at its call site, where the contracts are proven absent), which would destroy live contracts here. The new method only invalidates caches: the JDBC store invalidates its lookup cache, the in-memory store is a no-op because it has no cache.

**A persistent resolution failure now stalls the indexer deliberately.** The retry converges when the failure was transient (a stale cached mapping, or a ledger-end-cache window). If the activation reference is genuinely unresolvable, the indexer keeps restarting: we consider a loud stall preferable to silently persisting corruption that pruning later makes undiagnosable. If you would rather have an operator escape hatch (bounded retries, or a flag restoring the old warn-and-persist behavior), we are happy to add one — we did not want to grow the config surface unilaterally.

**Can a legitimate flow produce an unresolvable deactivation?** We believe not. Pruning only removes an activation row together with the deactivation that consumed it, so a live contract's activation row is never pruned; witnessed-only consuming exercises take the `lapi_events_various_witnessed` path and need no back-pointer. An unresolved reference at ingestion time is therefore either transient or evidence of an inconsistent index database. The integration test environments run `verifyIntegrity` at teardown, so a green CI run across the ACS-import, party-replication, and pruning suites is empirical evidence for this claim — we would appreciate maintainers double-checking the reasoning.

## How tested

- `community-common` / `ledger-api-core` (main and test) pass `scalafmtCheck` and compile.
- `ParallelIndexerSubscriptionSpec` extended: the throw from `dbPrepare` (both failure stages: internal-contract-id resolution miss and `lastActivations` miss), the invalidation contents, the single summarizing error log, the metric increment (via `InMemoryMetricsFactory`), the no-invalidation happy path, and the invariant violation through `batcher`.
- `StorageBackendTestsIntegrity` extended: the new check reports a NULL back-pointer below the ledger end and at exactly the pruning point, and ignores rows beyond the ledger end. One pre-existing test comment claimed NULL back-pointers are "not reported"; that test only stayed green because the stray-deactivation check fires first, and the comment now says so.
- Full local run: 304 tests, 0 failures across `ParallelIndexerSubscriptionSpec` and the complete `StorageBackendSpecH2` suite.

Not covered locally, deferred to CI: the Postgres storage-backend variant, downstream module compilation, and the integration suites. Two known coverage gaps we chose not to close here because they need the full e2e rig: the production wiring of the invalidation callback is verified by compilation only, and no test pins that the new exception restarts rather than halts the indexer (the sibling `ReferencedContractNotFoundException` has the same gap).

## User-facing changes

New metric, new integrity check message, and a behavior change (indexer restart instead of silent corruption) — documented in `UNRELEASED.md` under Bugfixes.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -206,6 +206,41 @@ The Ledger API command completion service now exposes a `GetCompletionByHash` en
 
 #### Recommendation
 
+### The indexer no longer persists deactivation events with an unresolvable activation reference
+
+#### Issue Description
+When the indexer could not resolve the activation reference of a deactivation event, it logged a
+warning and persisted the event with a NULL `deactivated_event_sequential_id`. Ledger API
+activeness queries anti-join on that reference, so the archived contract stayed visible as active
+in the ACS permanently, while ACS commitments (computed from the synchronizer-side stores)
+remained correct.
+
+#### Affected Deployments
+Participant nodes.
+
+#### Impact
+Participants hosting the same party could serve diverging ACS contents; aggregates computed from
+the ACS (e.g. a token registry's total supply) differed per node.
+
+#### Symptom
+A contract remains in `/v2/state/active-contracts` responses although the transaction stream shows
+its archival. The participant log contains "Activation is missing for a deactivation" warnings at
+the time the archival was indexed.
+
+#### Workaround
+Run the index database integrity check to detect affected rows; contact support for repairing the
+affected index database entries.
+
+#### Likeliness
+Very rare; requires a transient contract-resolution failure (e.g. a stale contract cache mapping
+or an indexer restart window) at the moment the archival is indexed.
+
+#### Recommendation
+Upgrade. The indexer now fails the ingestion batch and restarts instead of persisting the
+corruption, exposes the restarts via the new
+`daml.indexer.indexer_restart_due_to_unresolved_deactivation` metric, and the integrity check
+reports deactivation events without an activation reference.
+
 ## Deprecations
 - `StaticSynchronizerParameters.defaultsWithoutKMS` has been deprecated in favor of `StaticSynchronizerParameters.defaults`. Supported cryptographic schemes now have parity between KMS and non-KMS configurations.
 
--- a/community/common/src/main/scala/com/digitalasset/canton/participant/store/ContractStore.scala
+++ b/community/common/src/main/scala/com/digitalasset/canton/participant/store/ContractStore.scala
@@ -86,6 +86,13 @@ trait ContractStore extends ContractLookup with FlagCloseable {
     */
   def contractsPruned(internalContractIds: Iterable[Long]): Unit
 
+  /** Invalidate any cached lookups for the given internal contract IDs without modifying the
+    * underlying store. In contrast to [[contractsPruned]], this must never remove contract data: it
+    * is used when a cached mapping is suspected to be stale and a fresh read from persistence is
+    * needed. Stores without a cache do nothing.
+    */
+  def invalidateCachedContracts(internalContractIds: Iterable[Long]): Unit
+
   def contractCount()(implicit traceContext: TraceContext): FutureUnlessShutdown[Int]
 
   // TODO(i24535): implement this on db level
--- a/community/common/src/main/scala/com/digitalasset/canton/participant/store/db/DbContractStore.scala
+++ b/community/common/src/main/scala/com/digitalasset/canton/participant/store/db/DbContractStore.scala
@@ -470,6 +470,11 @@ class DbContractStore(
   override def contractsPruned(
       internalContractIds: Iterable[Long]
   ): Unit =
+    invalidateCachedContracts(internalContractIds)
+
+  override def invalidateCachedContracts(
+      internalContractIds: Iterable[Long]
+  ): Unit =
     internalContractIds.view
       .flatMap(cache.getIfPresentAuxKey)
       .foreach(cache.invalidate)
--- a/community/common/src/main/scala/com/digitalasset/canton/participant/store/memory/InMemoryContractStore.scala
+++ b/community/common/src/main/scala/com/digitalasset/canton/participant/store/memory/InMemoryContractStore.scala
@@ -141,6 +141,10 @@ class InMemoryContractStore(
       .flatMap(internalIds.remove)
       .foreach(contracts.remove(_).discard)
 
+  // The in-memory store has no cache in front of it: it is the authoritative store itself,
+  // so there is nothing stale to invalidate.
+  override def invalidateCachedContracts(internalContractIds: Iterable[Long]): Unit = ()
+
   override def lookupStakeholders(ids: Set[LfContractId])(implicit
       traceContext: TraceContext
   ): EitherT[FutureUnlessShutdown, UnknownContracts, Map[LfContractId, Set[LfPartyId]]] = {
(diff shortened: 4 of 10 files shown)


scolear:

scolear:
recheck

PR #594, opened 2026-07-27
test(external-call): end-to-end infrastructure and honest-path coverage

angelol:
Third slice of #564: the end-to-end test infrastructure and the honest-path scenario — the keystone the remaining #564 scenarios stack on.

## What changes

- **`CantonExternalCall` test package** (dev LF, `--target=2.dev`, mirroring `CantonLfDev`): an `ExternalCallTester` template whose nonconsuming `CallExtension` choice invokes `DA.ExternalCall.externalCall`. Wired into the Java codegen and exposed as `BaseTest.CantonExternalCallPath`.
- **`UseExtensionService` environment plugin**: runs a mock extension service over HTTP on a unique port and configures it on every participant under `parameters.engine.extensions`. It serves the external-call and version endpoints, records every call (mode, function id, config hash, body, request id, idempotency key) for test assertions, and answers with a fixed hex output unless a test installs its own responder — the affordance the follow-up failure scenarios (unreachable service, output mismatch) will use. Retries are disabled in the injected config so those scenarios observe a deterministic number of calls.
- **`ExternalCallIntegrationTest`** — the honest path end-to-end: upload the DAR, create-and-exercise the choice, and assert that the choice's result is the service output recorded in the transaction, and that the service saw exactly two calls — one in submission mode computing the output and one in validation mode re-validating it in the confirmation workflow — as distinct logical calls (distinct idempotency keys).

## Note on CI gating

The external-call wire data exists only at protocol version dev, so the suite uses the standard dev-gating (`testedProtocolVersion` + `onlyRunWithOrGreaterThan ProtocolVersion.dev`, per the no-hardcoded-PV rule from the #552/#567 reviews): it executes in the dev-protocol-version job on main and reports ignored in stable-PV PR jobs. If you'd prefer PR-time coverage for this suite, we can add a scoped dev job keyed on a test tag (`temp_acs_commitment_tests_pv_dev`-style) as a follow-up.

## Verified locally

Suite green at `CANTON_PROTOCOL_VERSION=dev` and reported ignored at stable; full `community-participant/test` green at both stable and dev protocol versions (the `BaseTest`/build changes touch everything); whole-repo `scalafixCheck` and `sbt format` clean.

Remaining #564 scenarios (disagreeing/missing outputs, unreachable service during submission/validation, recorded-vs-computed mismatch, V4-hashing prepare/execute) will stack on these fixtures as follow-up PRs.

PR #590, opened 2026-07-23
test(external-call): protocol-version gating and full alarm assertions in the protocol integration test

angelol:
First slice of #564: the two test cleanups agreed during the #567 review.

## What changes

- **Derive the protocol version via the usual dev-gating** ([#567 comment](https://github.com/digital-asset/canton/pull/567#discussion_r3535114765)): `ExternalCallProtocolIntegrationTest` no longer hardcodes `ProtocolVersion.dev` in its transaction factory. The suite now uses the default factory at the tested protocol version; the 14 tests that record external-call results (a dev-only wire field) are gated with `onlyRunWithOrGreaterThan ProtocolVersion.dev`, mirroring `ExternalCallConsistencyCheckerTest`. The no-results test and the five confirmation-response tests do not depend on the dev-only field and keep running at every protocol version — verified locally: 20/0 at dev, 6 passed + 14 ignored at stable.
- **Assert the full disagreement-alarm message** ([#567 comment](https://github.com/digital-asset/canton/pull/567#discussion_r3535130259)): the alarm assertions use `shouldBeCantonError` with the complete message ("Observed inconsistent external call results: …"), built from the same `Inconsistency` each test already expects in the per-view result, plus the `requestId` MDC as the context assertion. The two-mismatch test asserts both alarms in key order, which `validateRecordedResults` guarantees (its input is sorted by key and the traversal is order-preserving).

One commit per cleanup. The `MalformedRejects.MalformedRequest` code-only assertion in the response-factory section is intentionally untouched — the review ask was scoped to the disagreement alarm.

Closes the two test-cleanup checkboxes of #564 (the e2e/negative-path items remain).

PR #634, opened 2026-08-19
test(external-call): cover the failure scenarios end-to-end

angelol:
Part of #564: the "service unreachable during submission", "service unreachable during validation", and "recorded output != computed output" scenarios from the issue, plus two cheap siblings the fixture makes free.

**Stacked on #594** — the first four commits are #594 (including the `ExternalCallTest` rename); **only the last commit (b05938196) is new here.** I will rebase onto main as soon as #594 merges, leaving just that commit.

Extends `ExternalCallIntegrationTest` to six tests, each asserting the exact client-visible error, the participant's warnings and alarms (nothing more — the log windows are strict), and that a rejected request leaves nothing on the ledger:

- **Unreachable at submission**: the mock plugin now also configures an extension at a reserved-but-unbound port, so the call fails with a real connection refusal. The command is rejected synchronously with `INTERPRETATION_EXTERNAL_CALL_ERROR_EXECUTION_FAILED` carrying only the generic non-actionable client message, and no request reaches any service.
- **Non-canonical output at submission**: `INTERPRETATION_EXTERNAL_CALL_ERROR_INVALID_OUTPUT`, exactly one service call.
- **Service failure during re-validation**: the only confirmer abstains and the mediator rejects; the completion carries `CANNOT_PERFORM_ALL_VALIDATIONS` with the validator's status-only reason. A transport-level failure between the two phases of a single request cannot be injected with a static mock, but the equivalence is exact: the validator consumes only the error's status code, and a connection refusal maps to the same 503, so the abstain reason is byte-identical either way (the refusal path itself is covered end-to-end by the first scenario).
- **Non-canonical output during re-validation**: same abstain outcome, and the strict log assertion proves no warning and no alarm fires — pinning the deliberate asymmetry with the submission-time `INVALID_OUTPUT` reject.
- **Disagreeing re-validation output**: the `EXTERNAL_CALL_RESULT_DISAGREEMENT_ALARM` security alert fires exactly once per request with the full inconsistency description asserted by equality (output byte sizes only — the assertion proves the output bytes cannot leak into logs), and the completion carries `LOCAL_VERDICT_EXTERNAL_CALL_RESULT_DISAGREEMENT`.

Gating as in #594 and per the no-hardcoded-PV rule: `onlyRunWithOrGreaterThan ProtocolVersion.dev`, so the tests execute in the dev-protocol-version job and report ignored in stable PR jobs. Verified locally: 6/6 green at `CANTON_PROTOCOL_VERSION=dev`, 6 ignored at the stable default.

The remaining #564 scenarios (disagreeing outputs across participants, missing output, explicit V4 prepare/execute) need a second participant and the malicious-participant machinery and will follow separately on these fixtures.

### DIFF
--- a/community/app/src/test/scala/com/digitalasset/canton/integration/tests/extension/ExternalCallIntegrationTest.scala
+++ b/community/app/src/test/scala/com/digitalasset/canton/integration/tests/extension/ExternalCallIntegrationTest.scala
@@ -4,14 +4,21 @@
 package com.digitalasset.canton.integration.tests.extension
 
 import com.daml.ledger.api.v2.transaction_filter.TransactionShape.TRANSACTION_SHAPE_LEDGER_EFFECTS
+import com.daml.ledger.javaapi.data.{ExercisedEvent, Transaction}
 import com.digitalasset.canton.BaseTest
+import com.digitalasset.canton.console.CommandFailure
+import com.digitalasset.canton.discard.Implicits.DiscardOps
 import com.digitalasset.canton.extcall.java as M
 import com.digitalasset.canton.integration.plugins.{UseExtensionService, UseH2}
 import com.digitalasset.canton.integration.{
   CommunityIntegrationTest,
   EnvironmentDefinition,
   SharedEnvironment,
+  TestConsoleEnvironment,
 }
+import com.digitalasset.canton.logging.LogEntry
+import com.digitalasset.canton.participant.protocol.validation.ExternalCallValidationError
+import com.digitalasset.canton.topology.PartyId
 import com.digitalasset.canton.version.ProtocolVersion
 
 import scala.jdk.CollectionConverters.*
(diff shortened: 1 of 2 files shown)


angelol:
Appended c712204a6: covers the third and last canonical error id — non-hex call arguments fail the submission with `INTERPRETATION_EXTERNAL_CALL_ERROR_PREPARATION_FAILED` before any request reaches a service. Suite is now 7 tests, all green locally at `CANTON_PROTOCOL_VERSION=dev` and ignored at the stable default.

angelol:
Rebased onto main 57408d020 to pull in the CI fixes. New tip 050d92f50, content unchanged (range-diff clean). Scenario suite 7/7 at dev.

angelol:
Rebased onto main 86fa81fa2 (contains the merged #594 and #590). New tip 4b501c90e, content unchanged (range-diff clean). Scenario suite 7/7 at dev. With #594 merged, the PR now contains only its own commits.

simonmaxen-da (approved):
LGTM

PR #637, opened 2026-08-19
test(external-call): cover V4 hashing on the interactive-submission surface

angelol:
Part of #564: the explicit V4 prepare/execute scenario (deferred from the #549 review).

**Stacked on #594** — the first four commits are #594 (including the `ExternalCallTest` rename); **only the last commit (f428ba2d4) is new here.** I will rebase onto main as soon as #594 merges, leaving just that commit.

A new `ExternalCallInteractiveSubmissionIntegrationTest` on #594's fixtures, where an external party drives the flow, so the external-call results are covered by the party's own signature rather than only the submitting participant's:

- **Happy path pinned to V4**: `prepare` with an explicit `HASHING_SCHEME_VERSION_V4` echoes V4 back, and the prepared transaction's exercise node carries the recorded external-call results (all five fields asserted against the mock: extension id, function id, config, input, and the service output as raw bytes). Executing with the external party's signature commits the exercise (the result is the recorded output), and the service sees exactly one submission-mode and one validation-mode call.
- **V2 refused at prepare time**: preparing the same exercise with `HASHING_SCHEME_VERSION_V2` fails with `Cannot hash node with LF serialization version VDev using hashing scheme V2. Please use hashing scheme V4 or higher.` — and the call-mode assertion pins that interpretation (one submission-mode call) precedes the hashing failure, while nothing reaches the confirmation workflow.
- **V2 refused at execute time**: a V4-prepared transaction presented as V2 at execute fails with the same hashing error before submission, so the service again sees only the preparation-time call.

One design constraint worth flagging: externally-signed submissions must consist of a single view, so the suite creates the `ExternalCallTester` contract in the setup and prepares standalone exercises instead of reusing the create-and-exercise idiom of the sibling suite.

Gating as in #594 and per the no-hardcoded-PV rule: `onlyRunWithOrGreaterThan ProtocolVersion.dev`, so the tests execute in the dev-protocol-version job and report ignored in stable PR jobs. Verified locally: 3/3 green at `CANTON_PROTOCOL_VERSION=dev`, 3 ignored at the stable default.

### DIFF
--- a/community/app/src/test/scala/com/digitalasset/canton/integration/tests/extension/ExternalCallInteractiveSubmissionIntegrationTest.scala
+++ b/community/app/src/test/scala/com/digitalasset/canton/integration/tests/extension/ExternalCallInteractiveSubmissionIntegrationTest.scala
@@ -67,7 +67,6 @@ class ExternalCallInteractiveSubmissionIntegrationTest
                     aliceE.toProtoPrimitive
                   ).create.commands.loneElement
                 ),
-                includeCreatedEventBlob = true,
               )
               .getEvents
               .loneElement
--- a/community/app/src/test/scala/com/digitalasset/canton/integration/tests/extension/ExternalCallInteractiveSubmissionIntegrationTest.scala
+++ b/community/app/src/test/scala/com/digitalasset/canton/integration/tests/extension/ExternalCallInteractiveSubmissionIntegrationTest.scala
@@ -1,0 +1,187 @@
+// Copyright (c) 2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.digitalasset.canton.integration.tests.extension
+
+import com.daml.ledger.api.v2.event.Event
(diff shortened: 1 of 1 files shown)


angelol:
Appended d662ab506 from a fresh self-review pass: drops an unused `includeCreatedEventBlob = true` from the setup submit (nothing consumes the blob — the decode reads only standard created-event fields, and prepare resolves the contract from the local ACS). No behavior change; re-verified 3/3 at dev PV, 3 ignored at stable; format + scalafixCheck clean.

angelol:
Rebased onto main 57408d020 to pull in the CI fixes. New tip f3dfc964e, content unchanged (range-diff clean). Interactive suite 3/3 at dev.

angelol:
Rebased onto main 86fa81fa2 (contains the merged #594 and #590). New tip 17f2bdb9b, content unchanged (range-diff clean). Interactive suite 3/3 at dev. With #594 merged, the PR now contains only its own commits.

simonmaxen-da (approved):
LGTM

PR #638, opened 2026-08-19
test(external-call): reject tampered confirmation requests end-to-end

angelol:
Part of #564 (the external-call e2e + negative-path tests). Adds the two malicious-participant tampering scenarios (items A2 and A3 in the issue #564 tracking comment) plus the shared prerequisite that lets `TestSubmissionService` interpret external calls.

**Stacked on #594** (the honest-path keystone). Only the last **3** commits are new here; the first 4 belong to #594. This branch will be rebased onto `main` once #594 merges.

### Commits (newest last)
1. **interpret external calls in `TestSubmissionService`** — wires an injectable `ExternalCallHandler` (default `ExternalCallHandler.Unsupported`) into `TestSubmissionService`, resuming the interpreter's `Result.Need.ExternalCall` with `ExternalCallMode.Submission`, so a malicious submitter can record external-call results (closes the #564 TODO). Also fixes a latent config bug: `enableLfDev` set only the *engine* config to dev, but `Engine.checkAllowedDeps` reads the allowed LF versions from the *interpretation* config, so a dev package was rejected with a misleading engine-config-shaped error. `interpretationConfig` is now an `Option` defaulting to `.Dev` under `enableLfDev`. No caller passes `interpretationConfig`, so the change is source-compatible; `.Dev` widens the allowed versions over `.Default` (identical `contractStateMode`), so no stable-package suite can regress.
2. **reject a confirmation request with dropped results** (A3) — a malicious participant strips a view's recorded `externalCallResults` to empty; honest participants re-interpret, find no output to replay (`ExternalCallReplayMissing`), and reject with a failed model-conformance check. Size-only key fields are asserted verbatim; the extension service observes zero calls.
3. **reject a confirmation request with disagreeing results** (A2) — a malicious participant records two root views exercising the same external call, then rewrites one view's recorded output. Every confirming participant's consistency check spots the two occurrences disagreeing, alarms once (the full `Inconsistency` description is pinned), and rejects both recording views, so the request cannot be committed. The disagreeing key is excluded from re-validation, so the service is never contacted.

### Gating
The suite uses standard env-driven dev-gating (`onlyRunWithOrGreaterThan ProtocolVersion.dev`) per the no-hardcoded-PV rule from the #552/#567 reviews: external-call wire data exists only at protocol version dev, so it runs in the dev-PV job on `main` and reports "ignored" in the stable-PV PR jobs. A scoped tag-keyed dev job could add PR-time coverage as a follow-up if preferred.

Local gates: dev 2/2, stable 2-ignored; `sbt format` + whole-repo `sbt scalafixCheck` clean; full `community-app` test compile plus a malicious-`TestSubmissionService` H2 suite green at both PVs (shared-infra blast radius).

### DIFF
--- a/community/app/src/test/scala/com/digitalasset/canton/integration/tests/extension/ExternalCallTamperingIntegrationTest.scala
+++ b/community/app/src/test/scala/com/digitalasset/canton/integration/tests/extension/ExternalCallTamperingIntegrationTest.scala
@@ -207,8 +207,9 @@ class ExternalCallTamperingIntegrationTest
           LogEntryOptionality.Required -> modelConformanceReplayMissing("participant2"),
         )
 
-        // Validation fails during re-interpretation, before the extension service would be
-        // re-contacted, so the service observes no call at all.
+        // Re-interpretation replays recorded results instead of contacting the service, and the
+        // stripped views leave the external-call check nothing to re-validate, so the service is
+        // never contacted.
         extensionService.observedCalls shouldBe empty
 
         assertPingSucceeds(participant1, participant2)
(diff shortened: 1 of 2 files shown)


angelol:
Appended two small follow-ups from a fresh self-review pass, no behavior change: afbe4d99e corrects two comments that misstated the validation mechanism (validation-side re-interpretation is replay-only, and the external-call check skips on empty recorded results rather than being pre-empted by model conformance; the disagreement alarm fires once per request on every participant receiving the views, not only confirmers), and 69608c749 drops the now-dead `interpretationConfig` class-parameter default (`apply` always passes it explicitly). Re-verified: 2/2 at dev PV, 2 ignored at stable; format + scalafixCheck clean.

angelol:
Rebased onto main 57408d020 to pull in the CI fixes. New tip 7bcee255d, content unchanged (range-diff clean). Tampering + honest suites 3/3 at dev; whole-repo scalafixCheck clean.

angelol:
Rebased onto main 86fa81fa2 (contains the merged #594 and #590). New tip 6448f9a46, content unchanged (range-diff clean). Tampering suite 2/2 at dev. With #594 merged, the PR now contains only its own commits.

simonmaxen-da (approved):
LGTM

PR #644, opened 2026-08-25
feat(external-call): enable hashing scheme V4 on protocol version 36

angelol:
First slice of landing external calls in PV36 (per the Slack plan with @paulbrauner-da; September 2nd train). This slice is self-contained and independent of the serialization-version decision for external-call results, which the remaining slices depend on.

### What it does
- Moves the V4 row of `MinimumProtocolVersionToHashingVersion` from `dev` to `v36` (onwards semantics keep dev supported), mirroring the V3-to-v35 promotion (86cea8e80).
- Binds the `ProtoVersion(32)` codecs of `ExternalAuthorization` and `SubmitterMetadata` to v36. Their v31→v32 delta is exactly the `HASHING_SCHEME_VERSION_V4` enum value plus the Stable→Alpha marker, matching the pre-existing `ViewParticipantData` `ProtoVersion(32) -> v36` binding; codec selection at v35 and below is unchanged.
- Fallout: the `ExternalTransactionProcessorSpec` rejection test expects the v36 minimum and no longer registers at v36; the interactive-submission demo derives its schemes via `getHashingSchemeVersionsForProtocolVersion` (keeps its dev coverage, resolves the TODO on the exact-key lookup); the five ledger-API proto comments claiming V4 is dev-only are updated with `proto-data.yml`/`openapi.yaml` regenerated; `UNRELEASED.md` notes the promotion.
- Second commit restores the V4 hashing specification in `external_signing_hashing_algorithm.rst` (V4 section, V3/V4 tabbed exercise encoding incl. `external_call_results`, V3 wording fixes) that was merged in #549 and unintentionally reverted by the 2026-07-02 internal sync (ff509b03a), with the supported-schemes table saying v36 per this promotion.

### What it deliberately does not do
External-call results themselves remain gated (LF feature at 2.dev; `minExternalCallResults = VDev`; `ViewParticipantData.supportsExternalCallResults` at dev) — a v36 V4 flow never encounters a node it cannot hash. Those gates move in the next slices once the serialization version for external-call results below 2.dev is settled.

### Verification
- `CANTON_PROTOCOL_VERSION=36`: `ExternalTransactionProcessorSpec` (the ungated honor test now runs V4), `InteractiveSubmissionVersioningIntegrationTest` + `ExternalPartyOnboardingIntegrationTest` (external-party flows with V4 as the default scheme, 21 tests green), `SerializationDeserializationTest` (24/24, round-trips V4 through the v36 codecs), `HashingSchemeVersionSpec`.
[...]

PR #633, opened 2026-08-19
test(external-call): cover external calls in Daml Script

angelol:
Part of #564 (the recovered cross-repo item: Daml Script coverage agreed with @paulbrauner-da in the daml#23099 review, https://github.com/digital-asset/daml/pull/23099#discussion_r3473140101).

**Stacked on #594** — the first four commits are #594 (including the `ExternalCallTest` rename); **only the last commit (27442b32f) is new here.** I will rebase onto main as soon as #594 merges, leaving just that commit.

Adds a dev-LF Daml Script project `ScriptExternalCallTests` and a `DamlScriptExternalCallIT` suite backed by #594's `UseExtensionService` mock:

- `externalCallReturnsServiceOutput`: the recorded service output flows back as the choice result, asserted in-script and pinned as the exact JSON script result.
- `unknownExtensionFailsWithExternalCallError`: a call to an unconfigured extension surfaces to the script as the typed `ExternalCallError` submit error added by daml#23099 (`ExecutionFailed` with the extension and function ids pinned).

Notes:
- No sbt wiring is needed: `DamlScriptIT` builds the project with `dpm` at test runtime.
- Gating follows the sibling `DamlScriptPVDevLFDevIT`: `protocolVersionForTesting = ProtocolVersion.dev`, so the scripts run in the dev-protocol-version job and report ignored at stable PR jobs. Verified locally: 3/3 green at `CANTON_PROTOCOL_VERSION=dev`, 3 ignored at the stable default.
- The `ExternalCallTester` template is duplicated from `community/app/src/test/daml/ExternalCallTest` with keep-in-sync comments in both copies — the two projects are built by different build systems and cannot share sources.

PR #648, opened 2026-08-26
feat(external-call): promote the external-call LF feature to 2.4-staging

angelol:
Starts the LF promotion agreed with @paulbrauner-da / Roger on Slack (PV36 train, September 2nd): flips `featureExternalCall` from `Inclusive(v2_dev, v2_dev)` to `From(v2_4_1)`, i.e. the `EXTERNAL_CALL` builtin is accepted from LF **2.4-staging** onwards. The declaration in `project/DamlLfFeature.scala` generates both the Scala feature trait and `daml-lf-features.json` (which damlc extracts from the canton jar), so merging this is what starts the repo dance: release canton → import into damlc → release damlc → re-pin in canton. The proto comment and the two LF-spec availability notes move to `>= 2.4` per the staged-feature precedent (secp256k1 validate key).

Safe before the dance completes:
- the range **widens** — 2.dev keeps the feature, so every existing `--target=2.dev` fixture and DAR compiled by the currently pinned damlc is unaffected;
- a hand-crafted 2.4-rc1 package now decodes but is still rejected by the engine's allowed-language-versions check on any participant without `enableLfDev` (2.4 is in neither the stable nor early-access set), and the transaction-level gates on recorded external-call results (`SerializationVersion`/PV) are untouched — those move in the next slice, alongside #644.

Verified locally: `DecodeV2Spec` 66/66 (the suite iterates the feature range, so the 2.4-rc1 row flips reject→accept and 2.1–2.3 stay rejected), interpreter `ExternalCallTest` 11/11, engine `ExternalCallEngineTest` 3/3, `ExternalCallProtocolIntegrationTest` 20/20 at dev PV; `sbt format` + whole-repo `scalafixCheck` clean.


community/daml-lf/archive/src/main/protobuf/com/digitalasset/daml/lf/archive/daml_lf2.proto
```diff
@@ -391,7 +391,7 @@ enum BuiltinFunction {
 
   //SECP256K1_VALIDATE_KEY = 4001; // REMAINS HERE TEMPORARELY TO PARSE CHECKED IN DARS
 
-  EXTERNAL_CALL = 5001; // *Available in versions >= 2.dev*
+  EXTERNAL_CALL = 5001; // *Available in versions >= 2.4*
```
roger-bosman-da: Please change the position to no longer be `dev-only-so-super-high`. This will break the decoding, since the location of the external call will change, so we need to do a dance:
- make the canton change in the proto and disable all (but as little as possible) tests that deal with this temporarily
- issue a daml pr that ingest this updated canton and starts outputting according to updated proto
- ingest and re-enable tests here

please limit these "repo" dances to ideally one, so combine all changes that need to be danced in this manner

angelol: Done in ab26d8d95: `EXTERNAL_CALL` moved from 5001 to **74** (the enum's next stable id, per the `// Next id is 74.` marker, which is bumped to 75), entry relocated next to `SECP256K1_VALIDATE_KEY = 73`, and 5001 is `reserved`.

On the dance and test-disabling, the good news: **no tests on this branch need disabling.** Canton has no checked-in DARs using the builtin — the encoder-test fixtures (`Builtin_2.dev_.lf`) are re-encoded from source at build time, and the decoder/interpreter/engine suites are green with the new number (66/11/3). The breakage window only affects DARs emitted by a damlc that still encodes 5001, which today means (a) the daml repo's own external-call tests, handled by your ingest step, and (b) the damlc-compiled fixtures in the open external-call test-fleet PRs (#594 and the PRs stacked on it, #633). Those are unmerged, so sequencing is free to choose: merge the fleet after the dance completes and nothing ever breaks, or before it and their suites need a temporary disable during the window.

On bundling: this PR now carries both dance-requiring changes — the feature-range flip (the features.json damlc consumes) and this renumber — so one dance covers everything. The serialization-side sibling #649 keeps a staging-range field number (`external_call_results = 1002` in the transaction proto); that one is canton-internal at runtime (damlc never emits transactions), so I'll renumber it into the stable range in #649 directly — no dance needed.

community/daml-lf/archive/src/main/protobuf/com/digitalasset/daml/lf/archive/daml_lf2.proto
```diff
@@ -391,7 +392,7 @@ enum BuiltinFunction {
 
   //SECP256K1_VALIDATE_KEY = 4001; // REMAINS HERE TEMPORARELY TO PARSE CHECKED IN DARS
 
-  EXTERNAL_CALL = 5001; // *Available in versions >= 2.dev*
+  reserved 5001; // was the 2.dev staging position of EXTERNAL_CALL
```
roger-bosman-da: dev ids don't need to be reserved after usage

angelol: Dropped in c38623264 (and the matching reserved line in #649, a67a32089).

angelol:
Rebased onto main 57408d020 to pull in the CI fixes. New tip 09114564b, content unchanged (range-diff clean). DecodeV2Spec 66/66; a trailing-blank-line format fix is folded into the last commit.

Tbo-DA:
@angelol Can you rebase again please? I've pushed a fix for the compliance label issue. (even if it should not be blocking).

angelol:
Rebased onto current main (with the compliance-label fix) and fixed the red: the Protobuf Continuity job flags the intentional builtin renumber, so the snapshot is refreshed in bf843562c, matching how previous intentional proto changes handled it. buf checks pass locally. One note: #649 refreshes the same snapshot for its field renumber — whichever merges second gets a trivial rebase + re-refresh from me.

roger-bosman-da (approved):

PR #668, opened 2026-09-17
Fix defer instance of Speedy.Control

puffnfresh:
This previously just eagerly evaluated the deferred computation. We just need to have a constructor which suspends the deferred computation and then unwraps it in the existing tail recursive loop.


puffnfresh:

community/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
```diff
@@ -1296,9 +1299,10 @@ private[lf] object Speedy {
     final case class Complete(res: SValue) extends Control[Nothing]
     final case class Error(err: interpretation.Error) extends Control[Nothing]
     final case object WeAreUnset extends Control[Nothing]
+    final case class Suspend[Q](thunk: () => Control[Q]) extends Control[Q]
```
remyhaemmerle-da: Can you please move the suspend between Expression and Question, and pattern match inside the `run` method in the same order ?

remyhaemmerle-da (commented):
hanks for this, good catch. we'll accept it. 

We're also planning to move away from ContT in this code. As the fix is small, low-risk, and restores correct Defer semantics, so we'll merge it on that basis.

Could you please applied the small change I suggest ?

Issue #653, opened 2026-08-31
external-call: add release notes


roger-bosman-da:
When backporting https://github.com/digital-asset/canton/pull/648, I have added a stub for release notes in the UNRELEASED.md

When this lands in the open source repo, the release notes should be fleshed out (and backported) 

cc @angelol

PR #669, opened 2026-09-22
docs(external-call): flesh out the release notes on release-line-3.6

angelol:
Closes #653: expands the `external_call` stub in UNRELEASED.md with the feature description — what external calls are, the config path, submission-time recording and confirmation-time re-validation, the early-access requirements (LF 2.4, protocol version 36), and the hashing scheme V4 coverage for externally signed transactions.

Same text as #654, which targeted main before the 3.6 branch cut; if this is the right home for it now, #654 can be closed.


paulbrauner-da (approved):

paulbrauner-da:
Something is broken with the CI on release-line-3.6 in the digital-asset/canton repo. I will cherry-pick this change internally and we'll publish it back during the next drop.

